PR #12, opened 2022-10-14
Basic update of package.json

javagl:
Updated repository name and some dependencies.

There are still vulnerabilities (see https://stackoverflow.com/questions/68333071/how-to-solve-this-npm-glob-parent-problem if you want to tackle them), and `npm run test` does not work. But the basic installation should work now, and the actual tools that I tried out seemed to work as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,13 +9,13 @@
   "keywords": [
     "3D Tiles"
   ],
-  "homepage": "https://github.com/CesiumGS/3d-tiles-validator",
+  "homepage": "https://github.com/CesiumGS/3d-tiles-tools",
   "repository": {
     "type": "git",
-    "url": "https://github.com/CesiumGS/3d-tiles-validator.git"
+    "url": "https://github.com/CesiumGS/3d-tiles-tools.git"
   },
   "bugs": {
-    "url": "https://github.com/CesiumGS/3d-tiles-validator/issues"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,9 +25,9 @@
     "bluebird": "^3.5.3",
     "cesium": "^1.55",
     "fs-extra": "^7.0.1",
-    "gltf-pipeline": "^1.0.6",
+    "gltf-pipeline": "^3.0.5",
```
lilleyse, line 28: This might have to stay pinned to version 1.x.x for files like `optimizeGlb` to work. Really though the `2.0-tools` branch should be cleaned up and merged.

lilleyse:
Since packages have been updated could you regenerate `ThirdParty.json`?

javagl:
This was only an attempt to quickly get this to a state where compiling and certain parts of the tools can be used. I'll try to look closer into this (test failure, check of `optimizeGlb`, update of `ThirdParty.json`...) soon.

javagl:
I have pinned `gltf-pipeline` to a 1.x version, and updated `ThirdParty.json`. 

The build is still failing, and the tests cannot be run, but I consider this as part of https://github.com/CesiumGS/3d-tiles-validator/issues/7 and the task to integrate the [currently open branches](https://github.com/CesiumGS/3d-tiles-tools/branches).

PR #14, opened 2022-11-25
Do not remove .cmpt extension from the given output path.

wmatyjewicz:


javagl:
This has become obsolete with the update to 0.2.0. 

In the previous versions, the tools tried to make some guesses in terms of "default output file names". And of course I **do** see the point of convenience: Running `cmptToGlb -i example/file.cmpt` **without** having to add output `-o example/file.glb` would be nice and easy. 

But it raises a bunch of questions: 

- What should these "default names" be? 
  - For the example above, one could default to `example/file.glb` or to `example/file.cmpt.glb` (as of this PR). 
  - On top of that: One has to take into account that there may be _multiple_ output files for this operation...
- Should the default names be _specified_ as part of the command line interface? 
- Is changing the default name a 'breaking change'?
- ...

So in version 0.2.0, the output name is usually _mandatory_.

PR #16, opened 2023-02-13
TypeScript rewrite

javagl:
An initial draft of a possible 3D Tiles Tools rewrite. Further details will be added here soon.

---

Update (2023-04-12)

The previous update summarized the main part of the rewrite, with two further PRs still being pending. These PRs have been

- https://github.com/javagl/3d-tiles-tools/pull/1
- https://github.com/javagl/3d-tiles-tools/pull/2

They have been reviewed and merged into the `ts-tools` staging branch in the meantime. 

Similarly, the following, additional PRs have been created, reviewed and merged into the the staging branch: 

- https://github.com/javagl/3d-tiles-tools/pull/3
- https://github.com/javagl/3d-tiles-tools/pull/4

Both contain further, detailed descriptions of the changes. The README.md and IMPLEMENTATION.md in this branch have been updated to reflect these changes as well.

----

Previous updates:

Update (2023-03-14) : 

The main part here has reviewed, and the summary of that first review is in [the comment below](https://github.com/CesiumGS/3d-tiles-tools/pull/16#issuecomment-1456603894). But for many reasons, it probably makes sense to **not** hastily merge this PR into `main`, but instead, wait until certain anticipated features are implemented, some futher cleanups have been made, and there have been more tests on _actual_ data sets.

In the meantime, there are some PRs that are supposed to be added here, using the `tools-ts` branch as a sort of "staging" branch:

- https://github.com/javagl/3d-tiles-tools/pull/1
- https://github.com/javagl/3d-tiles-tools/pull/2

PR #17, opened 2023-03-01
Add command to analyze tile data

javagl:
**This is built upon the state from https://github.com/CesiumGS/3d-tiles-tools/pull/16 , and should only be merged after this PR has been merged!**

There are some tasks that regularly appear when debugging tileset data:

- Extracting the GLB data from a B3DM or I3DM
  - Often followed by wanting to extract the JSON part of that GLB...
- Extracting the feature- or batch table from  B3DM, I3DM, or PNTS
- Extracting the inner tiles from CMPT (and recursively doing the aforementioned operations)

This PR adds an `analyze` command. It receives an input **file** and an output **directory**, like so
```
npx ts-node ./src/main.ts analyze -i ./specs/data/contentTypes/content.b3dm-o ./output/analyzeTest -f
```

It then writes several files into the output directory. The file names are the input file name, with the following suffixes:

- `.batchTable.json`: Contains the batch table JSON
- `.featureTable.json`: Contains the feature table JSON
- `.layout.json`: Contains layout information (details below)
- `.glb`: Contains the GLB part of the input
- `.glb.json`: Contains the JSON of the GLB of the input

The command can be applied to B3DM, I3DM, CMPT, PNTS, and GLB (both glTF 1.0 and glTF 2.0), and will write the required files accordingly. (For composite tiles, the output files will carry an `.inner[i]` part, as in `content.cmpt.inner[1].batchTable.json`).

The `.layout.json` just contains start-, length-, and end-information about the elements in the tile, like this:
```
{
  "magic": "b3dm",
  "headerLength": 28,
  "byteLength": 9672,
  "featureTableJson": {
    "start": 28,
    "length": 92,
    "end": 120
  },
  "featureTableBinary": {
    "start": 120,
    "length": 0,
    "end": 120
  },
  "batchTableJson": {
    "start": 120,
    "length": 624,
    "end": 744
  },
  "batchTableBinary": {
    "start": 744,
    "length": 0,
    "end": 744
  },
  "payload": {
    "start": 744,
    "length": 8928,
    "end": 10416
  }
}
```
This can be useful for figuring out alignment issues or why certain parts are empty.

---

Open discussion points (beyond some aspects that are still a bit "drafty" right now): 

- It currently tries to write all files when they don't exist or `--force` was given, and prints a warning message when one file cannot be written. I think this makes sense for something that is primarily a debugging tool, but there may be other solutions.
- Right now, it is **formatting** all JSON output by default. This is useful for a quick glance at the JSON, but it should be configurable. It should be possible to write out the data **as-it-is**. Maybe even with separate files with suffixes like `.batchTable.json` and `.batchTable.formatted.json`...?


javagl:
The PR where the `anaylyze` command is added has been retargeted to the branch for the main PR: https://github.com/javagl/3d-tiles-tools/pull/1

PR #18, opened 2023-03-07
Support for implicit tileset traversal (and metadata handling)

javagl:
**This is built based on the state from https://github.com/CesiumGS/3d-tiles-tools/pull/16, and should only be merged after this PR is merged**

The `3d-tiles-validator` includes code for the traversal of implicit tilesets. Compared to the fairly [trivial traversal of _explicit_ tiles](https://github.com/javagl/3d-tiles-tools/blob/7339891572ff58c011967609b89acab57abca700/src/tilesets/Tiles.ts#L78-L85), a generic traversal of implicit tilesets entails quite some infrastructure:

- Subtree files have to be loaded. Availability information has to be extracted and made accessible. 
- There have to be low-level classes for Quadtree/Octree coordinates.
- Metadata values may _override_ certain tile properties. So a proper, generic traversal requires an infrastructure for metadata access.
- Subtrees may contain metadata in _binary_ form. This basically combines all of the above...

Code for all this already exists in the validator. Large parts of this code had just been ported from a private project to TypeScript, meaning that the structures already had been somewhat generic and independent of the task of validation (even though some _specific_ functionality had to be added for the validator). 

The goal of this PR is to carve these functionalities out, and offer them in a form that can be used for various operations that are part of the `3d-tiles-tools` functionality.

There are some "conceptual" differences that require some more work:

- The validator could perform many operations under the assumption that certain structures had already been validated, whereas in the generic case, more error handling will be necessary
- There's the question _which_ parts of this (large) chunk of infrastructure are supposed to be exposed on the API level
- There are currently basically _zero_ unit tests for the part that has been moved here, because most of the tests in the validator focussed on detecting whether the right validation issues have been reported, whereas the specs here will have to focus more on the right error handling and tests of the exposed parts of the API

The current state is a first pass of moving the code here. The "core" of the functionality that is supposed to be offered is shown in a [first draft of a demo](https://github.com/javagl/3d-tiles-tools/blob/d8ca024be6c5e78573d594ff0b9565b2ae2d20da/demos/TraversalDemo.ts#L43):
```
    await TilesetTraverser.traverse(tileset, schema, resourceResolver,
      async (traversedTile) => {
        const tile = traversedTile.asTile();
        // Do something with the `Tile` here...
        return true;
      },
      depthFirst
    );
```
It should be possible to just traverse a tileset, and receive each `Tile` (regardless of whether it's explicit or implicit), to perform operations on it. But many details still have to be sorted out.


javagl:
This is replaced by https://github.com/javagl/3d-tiles-tools/pull/2 , which targets the base branch of the main PR.

Issue #19, opened 2023-04-17
Set up CI with GitHub actions


javagl:
This should be easy, but opened to keep track of it: There should be a GitHub action for CI.

Most of the build infrastructure was taken from the 3d-tiles-validator, so it should mainly be a matter of copying the [validator CI file](https://github.com/CesiumGS/3d-tiles-validator/blob/0e43590d2f8c784137b630032043420c58e94ee4/.github/workflows/ci.yml) here.

lilleyse:
Fixed in https://github.com/CesiumGS/3d-tiles-tools/pull/26

PR #26, opened 2023-04-23
Add basic github workflow for CI

javagl:
This addresses https://github.com/CesiumGS/3d-tiles-tools/issues/19 . But there recently have been reports about TypeScript compilation failing in the validator at https://github.com/CesiumGS/3d-tiles-validator/issues/272 , and since this part of the code was _moved_ here (from the validator), the build will likely fail here as well. I could not (yet) reproduce the build failure locally, but will try to investigate it ASAP.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: CI
+
+on: [push, pull_request]
+
+env:
+  CI: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


javagl:
The comment at https://github.com/CesiumGS/3d-tiles-validator/issues/272#issuecomment-1518924102 describes the (likely) reason for the upcoming build failure, and a workaround. That workaround will/should probably be done in a dedicated PR, after this one is merged, to be sure that the workaround works... around.

PR #30, opened 2023-04-25
Create a "content" object for single contents.

javagl:
When applying things like `b3dmToGlb` as part of a content stage within a pipeline, the output used `"contents"` even when there only was a single content, and a `"content"` would have been sufficient. While this is "valid", it's certainly unexpected and was not intentional - it happened due to a missing `return`... 

I'm still thinking about how to cover this with a spec...

### DIFF
--- a/specs/BasicTilesetProcessorSpec.ts
+++ b/specs/BasicTilesetProcessorSpec.ts
@@ -13,6 +13,8 @@ import { Tile } from "../src/structure/Tile";
 
 import { TraversedTile } from "../src/traversal/TraversedTile";
 
+import { TilesetEntry } from "../src/tilesetData/TilesetEntry";
+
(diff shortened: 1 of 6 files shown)


javagl:
The specs now check the behavior here on two levels:

- The low-level checks for the `updateTileContent` function that actually does the update
- Whether this also has the intended effect when applying it to an actual tileset in a `BasicTilesetProcessor`

One could argue about the latter: The pipelines are supposed to be able to model arbitrary operations on tilesets. When the number of operations that are implemented based on pipelines increases, then a systematic way of testing the input+output of these arbitrary pipelines can become more tricky. But since the `BasicTilesetProcessor` is something that is used as a somewhat generic building block, increasing the test coverage here certainly makes sense.

Issue #37, opened 2023-05-05
Create "content" object for single contents in traversal code


lilleyse:
While testing out `ImplicitToExplicitDemo` I noticed that the output tileset uses "contents" instead of "content". This is similar to the issue that was fixed in https://github.com/CesiumGS/3d-tiles-tools/pull/30.

I think this would be a matter of updating `ImplicitTraversedTile::asFinalTile`, `ExplicitTraversedTile::asFinalTile`, and `ImplicitToExplicitDemo::buildExplicitHierarchy`.

PR #38, opened 2023-05-05
Always use 'content' for single content

javagl:
There are several operations that (vaguely) do something with tile content. The tile content may either be a single `tile.content` object, or an array, as `tile.contents`. The `TraversedTile` interface tries to hide this difference: It has a single method `getRawContents` that always returns an array (that may also be an empty or single-element array). 

Until now, when obtaining the "_JSON_ representation" of the `Tile` object from a `TraversedTile` (via `asRawTile` or `asFinalTile`), then this JSON object contained a `contents` array, even when there only was a single content.   

With this PR, this is changed: The tile object should have a `content` when there is only one content, and use the `contents` _only_ when there really are multiple contents.

Fixes https://github.com/CesiumGS/3d-tiles-tools/issues/37 

---

An aside: The `TraversedTile` has `getRawContents` and `getFinalContents` methods, both always returning an array. This latter is the one that returns the contents _after_ the semantic overrides have been applied. I'm not perfectly happy with these _two_ methods, and that might be refactored at some point. But there certainly shouldn't be `getRawContent`, `getRawContents`, `getFinalContent`, and `getFinalContents` methods that have to be called by clients in order to figure out whether there are _zero_, _one_ or _multiple_ contents. Right now, clients can just call `for (const content of t.getRawContents()) doSomethingWith(content);`, which is a tad more convenient.

### DIFF
--- a/demos/ImplicitToExplicitDemo.ts
+++ b/demos/ImplicitToExplicitDemo.ts
@@ -26,6 +26,7 @@ async function buildExplicitHierarchy(
   // and contents (which contains the template URI)
   if (traversedTile.isImplicitTilesetRoot()) {
     delete explicitTile.implicitTiling;
+    delete explicitTile.content;
     delete explicitTile.contents;
   }
 
--- a/src/tilesets/Tiles.ts
+++ b/src/tilesets/Tiles.ts
@@ -15,8 +15,8 @@ export class Tiles {
    * defined `tile.contents`, or an empty array, when the tile does
    * not have contents.
    *
-   * @param tile The `Tile`
-   * @returns The content URIs
+   * @param tile - The `Tile`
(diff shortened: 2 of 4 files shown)


lilleyse:
This looks good, though there's one last issue - `ImplicitToExplicitDemo` needs to be updated so that it deletes both `"content"` and `"contents"` from the root tile [here](https://github.com/CesiumGS/3d-tiles-tools/blob/9f0f16bb945117f200b0781011df8c6fddd86d1f/demos/ImplicitToExplicitDemo.ts#L29), otherwise the output will be incorrect:

image

javagl:
Sorry, you had mentioned the `ImplicitToExplicitDemo` in the issue, but that one slipped through. 
Updated.

PR #40, opened 2023-05-17
Experimental KTX support

javagl:
<strike>Builds upon https://github.com/CesiumGS/3d-tiles-tools/pull/39</strike>

Merged with a more recent state, now based on https://github.com/CesiumGS/3d-tiles-tools/pull/61


javagl:
This now includes the state from https://github.com/CesiumGS/3d-tiles-tools/pull/61 (so that one should probably be merged first), and a fix to resolve the main issue that prevented this from being merged: 

The [`KtxUtility`](https://github.com/CesiumGS/3d-tiles-tools/blob/ec211202094072b174c3dd32b214455defc5a22b/src/ktx/KtxUtility.ts) class offers the functionality that one would like to have in order to convert "any" input image data (PNG or JPG) into a KTX image, and that is 
```
const ktxBuffer = await KtxUtility.convertImageData(pngOrJpgBuffer, options);
```

Under the hood, this uses the Basis Encoder from https://github.com/BinomialLLC/basis_universal/tree/ad9386a4a1cf2a248f7bbd45f543a7448db15267/webgl/encoder , which is built into a JS+WASM bundle and included here in the `./src/ktx/external` directory. As described in the [`/src/ktx/external/README.md`](https://github.com/CesiumGS/3d-tiles-tools/blob/ec211202094072b174c3dd32b214455defc5a22b/src/ktx/external/README.md), the latest state of the Basis Encoder caused a runtime error when generating UASTC. This appears to be fixed with updated linker flags (details in the README.md). 

I tested this by creating a glTF asset with different encodings of the same texture. Here's an example with the `DuckCM.png` from the `Duck` sample model:

[DuckCompare.zip](https://github.com/CesiumGS/3d-tiles-tools/files/12514746/DuckCompare.zip)

Dragging-and-dropping this into https://gltf.report/ and selecting the 'Textures' tab shows an overview of the textures, their encodings, and their memory requirements:

SemiCesium DuckCompare

I also tested this with a 2048x2048 texture (although certain settings cause the encoding to become pretty slow then...), and it seems to work as expected.

javagl:
The last commit adds a convenience layer around the KTX encoding, so that it is possible to encode all textures in glTF tiles with something like
```typescript
import { TileContentProcessing } from "./src/tilesetProcessing/TileContentProcessing";
import { TileContentProcessorsTextures } from "./src/tilesetProcessing/TileContentProcessorsTextures";

import { KtxEtc1sOptions } from "./src/ktx/KtxEtc1sOptions";
import { KtxUastcOptions } from "./src/ktx/KtxUastcOptions";

async function runDemo() {
  const tilesetSourceName = "./data/input/tileset.json";
  const tilesetTargetName = "./data/output/tileset.json";
  const overwrite = true;

  // Create a `TileContentProcessor` that encodes textures
  // to KTX, with the given options for ETC1S or UASTC
  const etc1sOptions: KtxEtc1sOptions = {
    compressionLevel: 2,
    qualityLevel: 96,
  };
  const uastcOptions: KtxUastcOptions = {};
  const tileContentProcessor = TileContentProcessorsTextures.encodeToKtx(
    etc1sOptions,
    uastcOptions
  );

  // Process the tileset source, and write it to the tileset target,
  // applying each `TileContentProcessor` to all tile contents
  await TileContentProcessing.process(
    tilesetSourceName,
    tilesetTargetName,
    overwrite,
    tileContentProcessor
  );
}

runDemo();
```

The functionality for encoding textures may still be added at the command line.

Issue #5, opened 2018-05-03
To-do list for upgrade command


lilleyse:
Tentative changes for `1.0` that should be handled in the `upgrade` command

* Move batch table hierarchy to extension: https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/303
* Fix leading slash: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/issues/141
* CESIUM_RTC to RTC_CENTER: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/issues/142
* `asset.gltfUpAxis` removed
* Padding: https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/225
* gltf 1.0 to 2.0
* content.url changed to content.uri
* Upgrade glTF semantic `BATCHID` to `_BATCHID`

Progress is in https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/tree/2.0-tools

javagl:
This issue originally referred to the upgrade from "pre-1.0" to "1.0" tilesets. Some of the bullet points have therefore been addressed already, or are tracked in dedicated issues, or have become obsolete (the latter also insofar that `gltf-pipeline` does the actual upgrade of glTF data). 

The question of the _scope_ of the `upgrade` command is still relevant, though, and maybe now more than before, because it may now refer to upgrading 3D Tiles 1.0 to 1.1.

There is a (somewhat preliminary) set of [options in the `TilesetUpgrader` class](https://github.com/CesiumGS/3d-tiles-tools/blob/4bf71d2b5a1651787f851971422fe168aa26780e/src/tilesetProcessing/TilesetUpgrader.ts#L26) that show what the upgrade is _currently_ doing, and this includes some of the bullet points from above.

Beyond that, one could consider to extend the `upgrade` functionality to cover more of what is described in the [3D Tiles 1.0 to 1.1 migration guide for tile formats](https://github.com/CesiumGS/3d-tiles/blob/7b364caf1b22708e3e49d1ae814c3cd8518d278a/specification/TileFormats/glTF/MIGRATION.adoc).

javagl:
<sup>(This comment contains some `TODO`s - I'll edit and extend this as we move on...)</sup>

[...]

PR #41, opened 2023-05-22
Upgrade generalization

javagl:
This PR addresses the generalization steps of the `upgrade` command. This addresses https://github.com/CesiumGS/3d-tiles-tools/issues/5 , with the building blocks that are supposed to be implemented here being summarized in [this comment](https://github.com/CesiumGS/3d-tiles-tools/issues/5#issuecomment-1529058173). Further details will be added here when the PR is updated.


javagl:
The first steps for the migration will be
- convert the BatchID to `EXT_mesh_features`
- convert PNTS to GLB

For the conversion of the BatchID to `EXT_mesh_features`, I created some basic support for the `EXT_mesh_features` extension in `glTF-Transform`, but this is not (yet) part of this PR - it may have to undergo more tests and reviews.

For the conversion of PNTS to GLB, my current line of thinking is that there should be _some_ functionality to conveniently read positions, normals and colors from PNTS, and shove them into something that eventually generates a GLB. First drafts for that are in the (preliminary, raw, undocumented) [`TileFormatsMigration`](https://github.com/CesiumGS/3d-tiles-tools/blob/1d9352621db505ae9e99754946c11e3ed75364b6/src/contentProcessing/TileFormatsMigration.ts) class. The basic idea is that one can obtain iterators for the PNTS data as in
```
    const positions = PntsData.createPositions(featureTable, binary);
    const normals = PntsData.createNormals(featureTable, binary);
    const colors = PntsData.createColors(featureTable, binary);
```
These will always offer the positions+normals as 3 float values, and the colors as 4 float values, regardless of the quantization, oct-encoding or things like RGBA565.

These things can then be added to something that is currently called `GltfPointCloudBuilder`, as in
```
      for (const normal of normals) {
        pointCloudBuilder.addNormal(normal[0], normal[1], normal[2]);
      }
```
Actually creating the GLB data from that data could be done in a different class, but that may be decided later.

Some thoughts and doubts have been revolving around the question of the _types_ that are used: Should these be `number[]` arrays, or `[number,number,number]` arrays, or should there be a typedef for the latter, or should all this operate on `Cartesian3` to begin with...? All that is not yet sorted out. The current state is only a snapshot.

javagl:
The last update contains **DRAFTS** for some basic classes like `PointCloudReader` and `PointCloudWriter`, but there are a million degrees of freedom for the design, and many of them still have to be sorted out. The high-level functionality for converting PNTS to GLB that is implemented based on these classes so far is that...

- POSITION, POSITION_QUANTIZED
- NORMAL, NORMAL_OCT16P
- RGBA, RGB, RGB565
- CONSTANT_RGBA (global)

are extracted, and can be written into a GLB in an "un-encoded" form (i.e. all VEC3/VEC4 FLOAT attributes). Compressing that GLB in various forms should be doable with `glTF-Pipeline` as a standalone step, with little effort.

Support for BatchID will require an integration of the `EXT_mesh_features` extension, which is still WIP.

---

As something that is not directly related to this PR, but came up in this context: 

Previously, several functions worked with `IterableIterator` instances. But ... long story short, they messed that up: As the name suggests, `IterableIterator` is a mix of an `Iterable` and an `Iterator`, and while there are seemingly appealing reasons why something should be both, mixing these things does not make sense (at all - not even a little bit). Specifically, all these `IterableIterator` instances could only be iterated over _once_. 
Now, instead of working with `IterableIterator`, the implementation uses `Iterable`, and these `Iterable` instances are iterable _multiple times_. It may require more review and test coverage, but I'm sure that this will avoid _lots_ of confusion in the long run. It's probably a good idea to make sure that an `Iterable` is iterable...

javagl:
The previous commits focussed on implementing support for `EXT_structural_metadata` in glTF-Transform. It would probably have been better to extract this as an isolated task, and not implement it along that `update-generalization`. But the most important parts of the generalized upgrade - namely, converting the legacy tile formats into glTF+Extensions - _requires_ an implementation of these extensions. It could also have been possible to just manually twiddle the extension into the raw, untyped JSON. But having an "object oriented layer" on top of that might be more convenient for potential users (and maybe less error-prone ... for the users, because bugs will then rather be _our_ fault). An example of this abstraction layer for `EXT_mesh_features` is shown in [this demo](https://github.com/CesiumGS/3d-tiles-tools/blob/3dc2971ab170755906be7d76c6ea6637f88f1b57/demos/GltfMeshFeaturesDemo.ts), but the `EXT_structural_metadata` extension is _far_ (!) more complex, and will have to be aligned with the current `metadata/binary` implementation that is used for 3D Tiles.

javagl:
The functionality for converting PNTS to GLB (+extensions) is approaching a feature-complete state. It took a bit of effort to create the _infrastrucure_ for that, e.g. support of `EXT_structural_metadata`. But with this infrastructure, the actual conversion made some progress. 

In a local test, this was applied to the `pointCloudWithPerPointProperties.pnts` example from the CesiumJS Specs.

This file contains this batch table:
```
{
  "temperature": {
    "byteOffset": 0,
    "componentType": "FLOAT",
    "type": "SCALAR"
  },
  "secondaryColor": {
    "byteOffset": 4000,
    "componentType": "FLOAT",
    "type": "VEC3"
  },
  "id": {
    "byteOffset": 16000,
    "componentType": "UNSIGNED_SHORT",
    "type": "SCALAR"
  }
}
```
and this feature table:
```
{
  "POSITION": {
    "byteOffset": 0
  },
  "RGB": {
    "byteOffset": 12000
  },
  "POINTS_LENGTH": 1000
}
```

From this information, a metadata schema is generated:
```
{
  "id": "ID_batch_table",
  "name": "Generated from batch table batch_table",
  "classes": {
    "class_batch_table": {
      "name": "Generated from batch table batch_table",
      "properties": {
        "temperature": {
          "name": "temperature",
          "description": "Generated from temperature",
          "type": "SCALAR",
          "componentType": "FLOAT32",
          "array": false,
          "normalized": false,
          "required": true
        },
        "secondaryColor": {
          "name": "secondaryColor",
          "description": "Generated from secondaryColor",
          "type": "VEC3",
          "componentType": "FLOAT32",
          "array": false,
          "normalized": false,
          "required": true
        },
        "id": {
          "name": "id",
          "description": "Generated from id",
          "type": "SCALAR",
          "componentType": "UINT16",
          "array": false,
          "normalized": false,
          "required": true
        }
      }
    }
  }
}
```
This is then inserted into the glTF `EXT_structural_metadata` implementation, and used for wiring the attributes of a point cloud with the property attributes:
```
{
  "asset": {
    "generator": "glTF-Transform v3.3.1",
    "version": "2.0"
  },
  "accessors": [
    {
      "type": "VEC3",
      "componentType": 5126,
      "count": 1000,
      "max": [
        2.5,
        2.5,
        2.5
      ],
      "min": [
        -2.5,
        -2.5,
        -2.5
      ],
      "bufferView": 0,
      "byteOffset": 0
    },
    {
      "type": "VEC4",
      "componentType": 5121,
      "count": 1000,
      "normalized": true,
      "bufferView": 0,
      "byteOffset": 12
    },
    {
      "type": "SCALAR",
      "componentType": 5126,
      "count": 1000,
      "bufferView": 0,
      "byteOffset": 16
    },
    {
      "type": "VEC3",
      "componentType": 5126,
      "count": 1000,
      "bufferView": 0,
      "byteOffset": 20
    },
    {
      "type": "SCALAR",
      "componentType": 5123,
      "count": 1000,
      "bufferView": 0,
      "byteOffset": 32
    }
  ],
  "bufferViews": [
    {
      "buffer": 0,
      "byteOffset": 0,
      "byteLength": 36000,
      "byteStride": 36,
      "target": 34962
    }
  ],
  "buffers": [
    {
      "uri": "buffer.bin",
      "byteLength": 36000
    }
  ],
  "meshes": [
    {
      "primitives": [
        {
          "attributes": {
            "POSITION": 0,
            "COLOR_0": 1,
            "_TEMPERATURE": 2,
            "_SECONDARYCOLOR": 3,
            "_ID": 4
          },
          "mode": 0,
          "extensions": {
            "EXT_structural_metadata": {
              "propertyAttributes": [
                0
              ]
            }
          }
        }
      ]
    }
  ],
  "nodes": [
    {
      "mesh": 0
    }
  ],
  "scenes": [
    {
      "nodes": [
        0
      ]
    }
  ],
  "extensionsUsed": [
    "EXT_structural_metadata"
  ],
  "extensions": {
    "EXT_structural_metadata": {
      "schema": {
        "id": "ID_batch_table",
        "name": "Generated from batch table batch_table",
        "classes": {
          "class_batch_table": {
            "name": "Generated from batch table batch_table",
            "properties": {
              "temperature": {
                "name": "temperature",
                "description": "Generated from temperature",
                "type": "SCALAR",
                "componentType": "FLOAT32",
                "required": true
              },
              "secondaryColor": {
                "name": "secondaryColor",
                "description": "Generated from secondaryColor",
                "type": "VEC3",
                "componentType": "FLOAT32",
                "required": true
              },
              "id": {
                "name": "id",
                "description": "Generated from id",
                "type": "SCALAR",
                "componentType": "UINT16",
                "required": true
              }
            }
          }
        }
      },
      "propertyAttributes": [
        {
          "class": "class_batch_table",
          "properties": {
            "temperature": {
              "attribute": "_TEMPERATURE"
            },
            "secondaryColor": {
              "attribute": "_SECONDARYCOLOR"
            },
            "id": {
              "attribute": "_ID"
            }
          }
        }
      ]
    }
  }
}
```

The current state is still lacking cleanups and tests, tests, tests, tests, but it looks like the conversion itself is already working reasonably well.

javagl:
A first pass for extending the migration functionality to also convert B3DM files to glTF+metadata did not take tooo much effort. 

---

In fact, the greatest caveat was still related to PNTS, namely properly supporting the batch table properties that have been compressed with the `3DTILES_draco_point_compression`. While decoding the Draco-compressed _feature table_ did already work smoothly for a while, this extension caused some trouble for everything that was related to batch tables. Usually, the property values of a batch table are stored (as JSON or) in the batch table binary. The whole implementation assumed that, everywhere. Now, the `3DTILES_draco_point_compression` extension comes along, and says: "Your princess is in another castle" - i.e. "Your data is in a different table". The current solution is not very elegant, and should be cleaned up with this extension in mind. 

But it's generally a bit worrisome that it is literally impossible to write anything that consumes batch tables and that does **not** break when it encounters a Draco-encoded batch table. Seeing something like the [example](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_draco_point_compression#batch-table) with
```
    "Intensity": {
        "byteOffset": 0,
        "type": "SCALAR",
        "componentType": "UNSIGNED_BYTE"
    },
    "Classification": {
        "byteOffset": 0,
        "type": "SCALAR",
        "componentType": "UNSIGNED_BYTE"
    },
```
will cause every implementation to try and read these properties, both at `byteOffset:0`. And ... maybe that's where the data actually **is**? _There is not even a way to detect that as being 'wrong'!_  The implementation will deliver wrong results, or crash. Unless it knows that extension...

---

However, a "batch run" of the conversion function on all PNTS and B3DM files from the CesiumJS specs does seem to produce valid GLBs. The output still has to be verified further, and the functionality has to be routed as `convertPntsToGlb` and `convertB3dmToGlb` to the command line, but it looks like this could at least soon be in a merge-able state, at least as a preview feature.

javagl:
The latest commits contain a few cleanups and fixes from local tests, as well as a first, basic CLI.

The CLI consists of `convertPntsToGlb` and `convertB3dmToGlb` functions, as in 
```
npx ts-node ./src/main.ts convertB3dmToGlb -i ./batchedWithBatchTable.b3dm -o ./batchedWithBatchTable.glb
```

---

A few things are still open;

**1. Further integrating the functionality**

This functionality is not yet offered as a "pipeline content stage", but that's generally easy to add. Offering it as a part that is integrated into the general tileset `upgrade` command may involve a bit more thought, but I started some drafts for that. For example, some `upgradePntsToGlb`/`upgradeB3dmToGlb` options that should probably be _disabled_ by default initially, to not break any existing usages of the `upgrade` call, and because of other caveats (related to the next points)

**2. Questions about the options/configuration settings for the upgrade**

This refers to things like integrating the `RTC_CENTER` from the feature table into the root node of the GLB. This is the recommended approach from the migration guide, but ... may have some issues. The large translation from the `RTC_CENTER` may cause rendering issues (namely exactly those that are described in the blog post at https://help.agi.com/AGIComponents/html/BlogPrecisionsPrecisions.htm that suggests the `RTC_CENTER` approach as the solution...)

For example:

The following is the CesiumJS/specs `pointCloudNormals.b3dm` example, converted to GLB, once _without_ the `RTC_CENTER`, and once with the `RTC_CENTER`. It is shown in the BabylonJS debug view that visualizes the vertex normals:

Cesium RtcNormals

<sup>(Yes, I saw the second result first, and first thought that I messed up some conversion of the feature table data, but ... that's not the case. The difference here is _only_ that of the RTC)</sup>

(Outputs are attached as [pointCloudNormals.zip](https://github.com/CesiumGS/3d-tiles-tools/files/11995284/pointCloudNormals.zip) )

So the integration of the `RTC_CENTER` should probably something that can be configured with an option, but ... **when** it is integrated, it may cause trouble.

Another configuration setting could be: Whether "object-typed" values from the batch table should be converted into JSON-strings in the property tables. For example, the `batchedWithBatchTable.b3dm` specs file contains an `info` property that contains an object. The conversion currently has to serialize this to JSON, as in
```
Property table from batch table:
id | Longitude           | Latitude           | Height             | info                          | rooms
0  | -1.31968            | 0.698874           | 6.155801922082901  | {"name":"building0","year":0} | ["room0_a","room0_b","room0_c"]
1  | -1.3196832683949145 | 0.6988615321420496 | 13.410263679921627 | {"name":"building1","year":1} | ["room1_a","room1_b","room1_c"]
...

Creating structural metadata property table from
id | Longitude           | Latitude           | Height             | info                                  | rooms
0  | -1.3196799755096436 | 0.6988739967346191 | 6.155801773071289  | "{\"name\":\"building0\",\"year\":0}" | ["room0_a","room0_b","room0_c"]
1  | -1.319683313369751  | 0.6988615393638611 | 13.410264015197754 | "{\"name\":\"building1\",\"year\":1}" | ["room1_a","room1_b","room1_c"]
...
```
But that may or may not be desired...

---

**3. Tests, tests, tests....**

The complexity of possible conversions is considerable, and many things may go wrong. This raises questions about the testing approach (and error handling - but I guess it will have to boil down to a large `try...catch` initially). During the development, I mainly relied on some local "batch processing" of the CesiumJS/Specs files for PNTS and B3DM, veeeery detailed debug logging (of the JSON structures, tables, and glTF metadata info), and manual inspection of the resulting GLBs.

It is not yet clear what could be a sensible, sustainable testing approach here. Some thoughts:

- I started creating some actual unit test specs for _very_ basic building blocks. But many of the functions should not be considered to be a part of an API, and may change considerably in the next iterations, depending on _which_ configuration options are offered _where_ and _how_
- Some of the conversions are hard to carve out into actual unit tests, because they involve a complex mix of interdependent JSON- and binary data, and ... _generate_ the same complex mix (as GLB with metadata).
- One could consider to use the CesiumJS/specs files as inputs, and _at least_ run the glTF-Validator on the output. That could be doable with relatively little effort, but the validator does not cover the metadata extensions, and the glTFs are generated with glTF-Transform (and therefore usually do not contain certain forms of errors anyhow), so this _might_ have relatively little _use_ as well
- To avoid regressions, one could use the CesiumJS specs files as inputs, run the "batch processing" of the upgrade on them, and use the result as a 'golden' output (probably with `.gltf` output instead of `.glb`, to make it easier to spot _what_ changed, and whether it's relevant). The 'golden' approach still has all sorts of issues (yeah, that GitHub line endings stuff, and the repo size in general), but it might be a reasonable first shot here...

javagl:
The latest changes focussed on further integration and testing ... and smaller issues that have been discovered during the tests.

The `upgrade` command now accepts a `--targetVersion X.Y` argument (with `1.0` being the default).

For all versions:

- The asset version will be set to the target version
- Content that uses a 'url' will be upgraded to use 'uri'
- The 'refine' value will be converted to be in all-uppercase.

For 1.0:

- glTF 1.0 models in B3DM or I3DM will be upgraded to glTF 2.0.

For 1.1:

- The '3DTILES_content_gltf' extension declaration will be removed
- PNTS and B3DM content will be converted to glTF

The last one is the most critical one (more details below).

---

One issue that I stumbled over was (of course) the y-up-vs-z-up one. The point positions are stored in z-up. But glTF expects y-up. It was not entirely clear what's the best place or way to handle this. One could just insert a new transform in the root node of the glTF for this, or modify the point positions. For now, I'm modifying the point positions, but maybe one could make a case for handling that differently.

---

Regarding the tile format migration:

The tests for this are currently based on the CesiumJS `Specs/Data/Cesium3DTiles` and `PointCloud` tilesets. An initial draft of the tests only converted the tile content files individually. But I now decided to apply the full `upgrade` to these tilesets. So the tests will operate on these files, in the `specs/data/migration/` directory, and they will

- read the tilesets from `input/` 
- write the upgraded versions to `output/` 
- Write the glTF JSON parts to `output_gltf/`
- Compare these JSON parts to what is contained in `golden_gltf/`

(When there are differences, then these have to be examined, and potentially be commited as part of the `golden_gltf`).

As an additional layer for testing, I added a Sandcastle (and a README.md) in the [`migration` directory](https://github.com/CesiumGS/3d-tiles-tools/tree/69bc7e30bde2ee2af7c8d51136aee6410d4b24c1/specs/data/migration). When serving the `migration` directory (**after** running the tests - i.e. when the `output` directory was filled), then this sandcastle allows selecting the respective tilesets and compare the input and output:

Cesium Migration 0001

Yes, this GIF (intentionally) shows one of the cases that is "not working": The opacity information from the PNTS in the RGBA case _seems_ to be lost. (But on closer inspection, it is there in the GLB - so that might be a CesiumJS issue...). Another difference is visible e.g. for quantized positions, but they look like a plain precision issue due to the RTC_CENTER. Other differences will have to be analyzed in a similar manner, to see whether they are bugs in the migration, or in CesiumJS, or just something that cannot sensibly be migrated at all.

javagl:
The last commits have mainly been focussing on fixes for issues that had been identified by visually inspecting the migration results, using the sandcastle as described in the previous comment. This included

- Proper handling of alpha values in point clouds: When the point cloud _may_ contain alpha values, then the colors are checked and if they indeed **do** contain non-opaque values, then the alpha mode of the output GLB is set to `BLEND` (instead of the default `OPAQUE`)
- When the input point cloud is quantized (including oct-encoded normals as a special form of quantization), then the output should be quantized as well. There is no _direct_ correspondence to the `NORMAL_OCT16P` normals. But using glTF-Transform as a tremendously convenient black box here, the approach now is: If the input positions/normals are quantized, then the output will use `KHR_mesh_quantization` for the positions/normals, respectively, with 16-bit quantization for both.

A _very_ high-level summary of the additions of this PR is actually given by the [last commit with the update of the implementation notes](https://github.com/CesiumGS/3d-tiles-tools/pull/41/commits/2d35d19b7cd595dc53b774cbac07421aca3d0559), and this may be an "entry point" for some sort of review.

javagl:
The last commits focussed on cleanups, fixes from local tests, and an update pass for the demos.

I'll mark this as 'Ready For Review', because there shouldn't be any significant _functional_ additions be added to this state. (Minor updates of documentation (e.g. CHANGES.md) or fixes that fall out of further tests may be added, but no implementations of new features). In terms of the actual functionality and infrastructure, the main goals that have been outlined in https://github.com/CesiumGS/3d-tiles-tools/issues/5#issuecomment-1529058173 should be achieved with the current state of this PR.

Some points (e.g. upgrading I3DM to GLB+Extensions, Batch Table Hierarchies, or handling CMPT) are still open, but should be tackled as dedicated PRs. (This one already did grow a bit too large anyhow). Further steps would involve defining the pre- and post conditions and configuration options of the `upgrade` command more precisely, and increasing the test coverage based on the decisions there.

lilleyse:
I won't be able to give this as deep a review as I would like, but it seems like the last things we talked about offline have been ironed out (`BLEND` mode and quantized positions) so I'm good with merging this. Thanks for the huge amount of work here @javagl.

lilleyse:
Any doc changes like updating `CHANGES.md` could happen prior to publishing a new version to npm.

PR #49, opened 2023-07-05
Apply RTC_CENTER from tile data to GLB

javagl:
The `b3dmToGlb` and `i3dmToGlb` operations until now only extracted the GLB payload from the tile data. The resulting GLB did not take into account any `RTC_CENTER` that may have been defined in the feature table of the input data.

This PR changes that: When there is an `RTC_CENTER` in the feature table, then this will cause new root nodes to be inserted into the GLB, which have the `RTC_CENTER` as their `translation` (taking the y-up-to-z-up transform into account). 

Some details to be decided:

**1.** Should applying the RTC_CENTER **always** be done? 

The `b3dmToGlb` and `i3dmToGlb` can either be used as content stages in a pipeline, or used as "standalone" commands from the command line. I currently assume that the RTC_CENTER should **always** be applied, but wonder whether someone might want an option to only extract the GLB, without applying the `RTC_CENTER`. Something like this could be offered with an option, e.g. by having a `-applyRtc=true/false` command line option that is `true` by default...

**2.** Should applying the `RTC_CENTER` also be done when extracting the (multiple) GLBs from a CMPT? 

The `cmptToGlb` command line operation takes one CMPT, and extracts **all** GLBs (recursively through other CMPTs or I3DMs or B3DMs). I assume that the results should also include the `RTC_CENTER` for now.

### DIFF
--- a/specs/ContentOpsSpec.ts
+++ b/specs/ContentOpsSpec.ts
@@ -1,0 +1,31 @@
+import fs from "fs";
+import { Buffers } from "../src/base/Buffers";
+
+import { ContentOps } from "../src/contentProcessing/ContentOps";
+import { SpecHelpers } from "./SpecHelpers";
+
+describe("ContentOps", function () {
+  it("extracts a single GLB buffers from B3DM", async function () {
+    const p = "./specs/data/contentTypes/content.b3dm";
+    const tileDataBuffer = fs.readFileSync(p);
+
+    const glbBuffers = await ContentOps.extractGlbBuffers(tileDataBuffer);
+    expect(glbBuffers.length).toBe(1);
+  });
+
+  it("extracts multiple GLB buffers from CMPT", async function () {
+    const p = "./specs/data/contentTypes/content.cmpt";
+    const tileDataBuffer = fs.readFileSync(p);
+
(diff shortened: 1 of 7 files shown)


lilleyse:
I think it's fine if `RTC_CENTER` is applied by default, with an option to override that. Same for inner cmpt contents as well.

javagl:
There is no option to _not_ apply it right now, but I can add that.
(Converting to draft for now)

javagl:
After some internal discussion, we decided to _not_ apply the `RTC_CENTER` to the GLB data by default.

The reasoning is that the `b3dmToGlb` command should primarily _extract_ the data, as it is. (It will still perform an attempt to upgrade from glTF 1.0 to glTF 2.0. But for glTF 2.0 assets, the asset should be extracted in an "unmodified" form)

There are many things that could go _beyond_ just extracting the data, and integrating the `RTC_CENTER` into the GLB would be only one of them. So this should be covered, together with maaaany other things, with the [generalized upgrade](https://github.com/CesiumGS/3d-tiles-tools/pull/41). Specifically: There will probably be a `convertB3dmToGlb` function that does ~"the things that go beyond extracting the data" (and what that is, exactly, still has to be sorted out...)

Issue #47, opened 2023-06-27
How to generate tileset.json


linglom1:
When I am using the command line to convert .gltf models to .b3dm, it does not generate a tileset.json file. How can I generate a tileset.json file?

javagl:
There currently is no functionality for generating a `tileset.json` file for glTF models.

(Note that with 3D Tiles 1.1, you don't have to convert the glTF asset to B3DM any more. You can use glTF directly as the tile content). 

However, it would be relatively easy to create such a functionality, and we'll consider to implement this in the near future.

There are some degrees of freedom for how "powerful" this functionality should be. If the goal is only to create a minimal `tileset.json` for a _single_ glTF asset, then the structure of this `tileset.json` would be simple:
```
{
  "asset" : {
    "version" : "1.1"
  },
  "geometricError" : 1.0,
  "root" : {
    "boundingVolume" : {
      "box" : [ 0.5, 0.5, 1.0, 0.5, 0.0, 0.0, 0.0, -0.5, 0.0, 0.0, 0.0, 1.0 ]
    },
    "geometricError" : 0.0,
    "refine" : "REPLACE",
    "content" : {
      "uri" : "content.glb"
    }
  }
}
```
The main purpose of offering the creation of such a JSON file as part of the tools would be to **compute the proper bounding box**. In the example above, the bounding box is just a unit cube. In order to define the bounding box so that it matches the one of the glTF file, one could use a function similar to [this example in the `BoundingBoxTests` sample](https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.1/BoundingBoxTests#conversion) to compute the tileset bounding box from the glTF bounding box.

(If the goal is only to do this once, then you could obtain the bounding box of the glTF with https://gltf.report/ and use this function to compute the bounding box that has to be put into the `tileset.json`)

linglom1:
Thank you very much for your response. My idea is to convert an FBX model into multiple GLTF models, and then generate a suitable tileset.json. I am really looking forward to the arrival of the function to automatically generate tileset.json. Finally, I wish you a happy life.

javagl:
> into multiple GLTF models, and then generate a suitable tileset.json

[...]

PR #51, opened 2023-07-22
Generate tileset JSON from content files

javagl:
This was requested in https://github.com/CesiumGS/3d-tiles-tools/issues/47 , and I think that it could be useful to have this functionality. 

<sup>(In fact, I already had to do this manually for some of the tests, and the `TODO Create function to create "dummy" tileset from GLBs` has been somewhere near the bottom of my TODO list for quite a while now)</sup>

This adds a `createTilesetJson` command that creates a `tileset.json` file for one or multiple GLB files. 

To create a `tileset.json` where the root tile has a single (`input.glb`) as its tile content:
`npx ts-node ./src/main.ts createTilesetJson -i C:/input.glb -o C:/tileset.json`

To create a `tileset.json` where the root node has children that refer to _all_ `.glb` files in the input directory (recursively):
`npx ts-node ./src/main.ts createTilesetJson -i C:/inputFiles -o C:/tileset.json`
(the content URIs will then be relative to the given `inputFiles` directory)

---

The bulk of the work here is computing the proper bounding volumes, including the computation of the bounding boxes of the glTF assets (done with glTF-Transform), computing the union of the bounding boxes for multiple inputs, and dealing with the y-up-to-z-zp transform.

I ran tests with the https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.1/BoundingBoxTests inputs, and these will probably become the basis for unit tests later.

Until now, this only supports `.glb` files. Generalizing this to other tile content types will involve different amounts of effort. 

- B3DM should be trivial: Just extract the GLB and use this
- I3DM may be a bit of a hassle. It might make sense to first implement the "I3DM-to-GLB+Extensions" `upgrade` functionality, and then hope that glTF-Transform delivers the right bounding volume for a GLB with `EXT_mesh_gpu_instancing`...
- PNTS should be easy: Just iterate over all point positions to compute the bounding box
- CMPT should be easy as soon as all others are covered

In all cases, there's that `RTC_CENTER` that may cause confusion (and no matter how hard I'll think about this, I know that the first solution will be _wrong_, but then, there should only be one _right_ solution left...)


javagl:
I found a GLB file where it does not compute the right bounding volume. Even though it does compute it correctly for the `BoundingBoxTests`, further cases will have to be checked, maybe based on the glTF sample models.

javagl:
A small update: The case where the bounding volume was not computed correctly was caused by https://github.com/donmccurdy/glTF-Transform/issues/879 - so this is not going to be addressed here.

javagl:
Beyond that, as a test, and ... for the heck of it, I just did a batch run of that over all binary glTF sample models, and the bounds seem to be right in all cases, except for the ones that involve skinning and instancing (and these are a different story anyhow)


Cesium 107 all glTF models

javagl:
We decided that it could be worthwhile to not only support GLB, but generalize this a bit to other "common" content types. These types will likely include B3DM, PNTS, I3DM and CMPT. The main work for generalizing it is in the bounding volume computations:

- For B3DM, it will be trivial: Just use the bounding volume of the contained GLB. (Caveats with the `RTC_CENTER`, as usual...)
- For PNTS, it should be simple: The `PntsPointCloud` class already offers an `Iterable<number[3]>` for the positions
- For I3DM, it may be tricky. One approach would be to compute the min/max of the GLB, and transform this with each instancing transform. This will of course not be the _tightest_ bounding volumes, but that's not even the goal here
- For CMPT, it should be relatively easy again, once the previous ones are all implemented

The current state supports GLB and PNTS (but is still work in progress)

javagl:
The latest updates include support for I3DM and CMPT as the input. 

- The PNTS support was trivial, by just computing the bounding box of all points.
- The I3DM support involved the usual hassle of Y-up-vs-Z-up, but since the I3DM matrices had been available conveniently from the `TileTableDataI3dm` class that was added in https://github.com/CesiumGS/3d-tiles-tools/pull/52 , this wasn't so much of an issue.
- The CMPT support was simple, with minor (internal) "refactorings" to make it possible to apply the existing functions for computing the bounding boxes of different content types to the inner tiles.

Two larger TODOs remaining for this PR:

1: Tests. 
I think that one reasonable approach here would also be that of a "golden output" approach: Using different contents from the CesiumJS specs, creating the `tileset.json`, and checking these into the repo as the reference that the output is compared against during the actual test run.

2: Better bounding volumes.
As pointed out in https://github.com/CesiumGS/3d-tiles-tools/issues/47#issuecomment-1684036368 , the bounding volumes that are computed for the tile content are by far not the tightest bounding volumes that _could_ be computed _threoretically_. One option that I considered was to compute the _oriented_ bounding box from the vertices in the input. But from a quick test, using [`OrientedBoundingBox::fromPoints`](https://cesium.com/learn/cesiumjs/ref-doc/OrientedBoundingBox.html#.fromPoints) did not return a proper bounding volume:

Cesium OBB

This might be related to the fact that there are _many_ points in the input (and the method might only be suitable for "fewer" points), but I haven't investigated the details here (yet).

Issue #53, opened 2023-08-01
Convert command should check for tileset.json or rename a supplied .json file


r-veenstra:
When using the convert command to convert a 3d tiles file based dataset to a .3dtiles database, the convert command should check either:

a) A tileset.json exists in the supplied folder before converting
b) If the user supplies a specific path a to a tilesets json (eg. example.json) it should rename to tileset.json on conversion

Otherwise, the user is able to generate a .3dtiles dataset that does not include a tileset.json.  When uploaded to hosting services such as Cesium ion, this tileset is not able to be used as the expectation is a tileset.json is present.

javagl:
When the input directory contains a `tileset.json` and a `mainTileset.json`, and the user gives `directory/mainTileset.json` as the input, then the `mainTileset.json` can **not** be renamed to `tileset.json`. 

That sounds like an unlikely case, but people can create new tilesets with the `merge` command, and that might well result in a "top-level" tileset JSON that is called `mergedTileset.json`. That one may _refer_ to the `tileset.json`, directly or indirectly, so it's not possible to simply rename that `tileset.json` either.

My gut feeling is that this is still unlikely enough to warrant an error message. Unfortunately, there would hardly be a way for the user to deal with this situation. A workaround - iff this case ever comes up - could be to "wrap" the tileset into another one, so that the structure of

`/directory/mainTileset.json`
`/directory/tileset.json`

would become

`/directory/tileset.json` (the top-level one that refers to `subdirectory/mainTileset.json` as an "external" one)
`/directory/subdirectory/mainTileset.json`
`/directory/subdirectory/tileset.json`

javagl:
I created a PR for addressing this.

The behavior for these cases that are addressed with this PR is summarized here:

There is a `basic` input (directory or ZIP file). It contains a `tileset.json` and a `tilesetA.json`. The conversions for the basic case:

- `basic/` to `basic.3tz`: Works. There is a `tileset.json` in the input
- `basic/tileset.json` to `basic.3tz`: Works. 
- `basic/tilesetA.json` to `output/resultTilesetA.json`: Works. (The `tileset.json` does not matter here)
- `basic/tilesetA.json` to `basic.3tz`: **Fails**. The `tilesetA.json` has to be renamed to `tileset.json` to be written into the 3TZ file, causing a duplicate with the other `tileset.json`
- `basic.zip` to `output/`: Works. This is just unpacking the ZIP.
[...]

PR #54, opened 2023-08-01
Handle tileset JSON file names for conversion

javagl:
This may fix https://github.com/CesiumGS/3d-tiles-tools/issues/53

The `convert` command did not enforce the requirement of a `tileset.json` file being present when the output was a `.3tz` or `.3dtiles` file. There are some corner cases where it is difficult to enforce this. Some details are discussed in https://github.com/CesiumGS/3d-tiles-tools/issues/53#issuecomment-1660667472


javagl:
The comment at https://github.com/CesiumGS/3d-tiles-tools/issues/53#issuecomment-1660667472 summarizes a few cases of possible inputs and outputs. The state that was described there was not sufficient yet, because there was no way to resolve the cases that had been marked with **"Fails"**. 

Specifically: When someone had a ZIP file that contained tilesets like `mainTileset.json` and `someOtherTileset.json`, then there was no way to convert this (e.g. into a `.3dtiles` file). We considered different options to resolve this. One idea was to check if the input ZIP file only contained a single `.json` file, and assume that this must be the Tileset JSON file. But that would be brittle and potentially confusing.

The approach that is implemented now:

- When the input is a full path to a JSON file, like `/example/mainTileset.json`, then the `mainTileset.json` is the input tileset JSON file name
- In all other cases (inculding the case where the input is a ZIP file), the input tileset JSON file name defaults to `tileset.json` (some archeology: https://github.com/CesiumGS/3d-tiles/issues/184 )
- In all cases, it is optionally possible to "override" the name of the input tileset JSON file, using the `inputTilesetJsonFileName` command line option

For example, consider an `ambiguous.zip` that contains a `tilesetA.json` and a `tilesetB.json` (as used in the specs).

The call
`npx ts-node ./src/main.ts convert -i ./specs/data/convert/ambiguous.zip -o ./output/ambiguous.3tz -f`
will cause the error
> Error: File specs\data\convert\ambiguous.zip did not contain a tileset.json

The call
`npx ts-node ./src/main.ts convert -i ./specs/data/convert/ambiguous.zip -o ./output/ambiguous.3tz --inputTilesetJsonFileName tilesetA.json -f `
will work. The `tilesetA.json` from the input will become the `tileset.json` in the output.

(Using `--inputTilesetJsonFileName notThere.json` will cause an error, because the ZIP does not contain this file)

PR #55, opened 2023-08-08
Handle package entry compression

javagl:
There are some details about the compression of the entries of 3D Tiles packages (`.3dtiles` and `.3tz` files) that are not yet fully specified, and where the implementation may have to be updated, based on the decisions that are made for the specification.

This PR covers a tiny detail that came out of a real-world example:

A 3TZ file _may_ contain entries that are compressed. And they may be compressed with ZSTD or with DEFLATE. For now, this PR changes the behavior of the 3TZ-based `TilesetSource` implementation: When it encounters entries that are compressed with DEFLATE, then they will be uncompressed _transparently_.

Whether or not this should also happen for ZSTD- or GZIP compressed entries is not yet decided.

### DIFF
--- a/demos/PackageSandcastle.js
+++ b/demos/PackageSandcastle.js
@@ -1,54 +1,32 @@
 /* eslint-disable no-undef */
 const viewer = new Cesium.Viewer("cesiumContainer", {
   globe: false,
-  skybox: false,
+  skyBox: false,
 });
 
 const tileset = viewer.scene.primitives.add(
-  new Cesium.Cesium3DTileset({
-    url: `http://localhost:8003/tileset.json`,
+  await Cesium.Cesium3DTileset.fromUrl(`http://localhost:8003/tileset.json`, {
     debugShowBoundingVolume: true,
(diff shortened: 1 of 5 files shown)


javagl:
This PR covers two things:

- Handle the case that entries of `.3tz` files are compressed with `DEFLATE`. These entries will be uncompressed transparently, and the corresponding method will always and only return the _uncompressed_ data.
- Fix some issues for large ZIP files in the "legacy" `ArchiveFunctions3tz` class. The latest state there did not properly handle certain data layouts for the `extras` block of ZIP entries (details from the spec added as [a comment](https://github.com/CesiumGS/3d-tiles-tools/pull/55/files#diff-74927cee6b096c1e83c57b6d0049c09ab311a44d6d86d8069272fae8378390daR89)). Also, some method used the `number` type where it is safer to use `bigint` - so now the implementation is using `bigint` wherever possible.

Some details about the handling of compressed entries in packages are not yet defined. I tried to summarize them in https://github.com/CesiumGS/3d-tiles-tools/issues/56 .

PR #57, opened 2023-08-27
Consolidation and cleanups

javagl:
(Builds on https://github.com/CesiumGS/3d-tiles-tools/pull/51 , so that has to be merged first)

This is a consolidation and cleanup PR. The previous release has been a while ago. There have been several PRs merged in the meantime. Some (minor) TODOs and maintenance work has been accumulated, and this is addressed here:

- Much of the functionality that had been created for the upgrade/migration had been created in the `contentProcessing/pointClouds` package initially and then been moved to `migration`. Much of this consists of functions for accessing the data of batch- and feature tables in a generic form. So it was moved to `tileTableData` now.
- The glTF metadata implementations carried a TODO that the (internal!) types could be restricted from `string` to `"UINT8" | "UINT16" ...`. This would not bring any runtime safety, but has been updated now to explicitly define the valid strings as the type.
- The 'gltfMetadata' package contained glTF extension implementations for glTF-Transform, including `EXT_instance_features`. So it has been renamed to `gltfExtensions`
- Addressed minor TODOs related to cleanups (TSDoc, string outputs...)
- Updated files for API-extractor: Mark everything as `@internal` that was not marked yet
- Updated ThirdParty.json
- Updated README.md, the CLI help, and CHANGES.md based on the latest changes (mainly for the `convert` and `upgrade` commands)


javagl:
The latest commits actualy fixed two bugs that had remained unnoticed until now:

- There was an `await` missing in the `TilesetUpgrader`, which could cause errors in nested tilesets
- There was a defined-check missing in the `GltfUtilities` that caused an error for glTFs that did not use extensions

The missing `await` caused me to temporarily/locally enable stricter linting rules (<strike>`recommended-type-checked`</strike> `recommended-requiring-type-checking`), and this revealed a few more missing `await`s, some unnecessary type assertions, and other minor linting fixes. We should consider enabling some of them by default (enabling all of them causes ~700 errors, many of them not really reasonable...), but this is beyond the scope of this PR.

PR #75, opened 2023-10-21
Maintenance 2023 10

javagl:
The `index.ts` currently exports _most_ of the classes of the 3D Tiles tools.

(Even though **all** classes are considered to be `@internal` for now. The tools are only available as a command-line application, with no public API. Eventually, the tools should be split into multiple packages, each with a sensible public API. This process has already been started in https://github.com/CesiumGS/3d-tiles-tools/pull/64 , but there is no fixed timeline (yet) for this conversion. In the meantime, the tools should export _everything_, for internal experiments (which also serve the purpose of fleshing out _what_ actually should be exported as a public API, eventually)).

This PR updates the `index.ts` to export all classes (except for the `pipeline` classes).  

In addition to the usual maintenance steps (update of the auto-generated API file and `ThirdParty.json`), this PR affects many files in the `TSDoc updates` commit. But this commit does **not** affect anything except for the TSDoc. It only adds missing TSDocs, hyphens in `@param example - There must be that hyphen` declarations, and missing `@internal` tags for classes.

### DIFF
--- a/ThirdParty.json
+++ b/ThirdParty.json
@@ -80,6 +80,14 @@
     "url": "https://www.npmjs.com/package/gltfpack"
   },
   {
+    "name": "meshoptimizer",
+    "license": [
+      "MIT"
+    ],
+    "version": "0.19.0",
+    "url": "https://www.npmjs.com/package/meshoptimizer"
+  },
+  {
     "name": "minimist",
     "license": [
       "MIT"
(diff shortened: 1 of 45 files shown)


javagl:
This also fixes https://github.com/CesiumGS/3d-tiles-tools/issues/77 : The `BasisEncoder` that is created in the `KtxUtility` class is now deleted by calling `delete()` on the implementation object when it is no longer used. I "verified" that this prevents the crash insofar that I did use it to consecutively encode hundreds of (relatively large) images in a single run, and it did not crash any more.

PR #86, opened 2023-11-19
Extended package server demo

javagl:
The current [`PackageServer` demo](https://github.com/CesiumGS/3d-tiles-tools/blob/194669391c1fd0476bc259b87d5c7e2911e86950/demos/PackageServer.ts) is a **very** basic implementation of a server that can serve data directly from 3TZ or 3DTILES packages.

This PR tracks the process of eventually making this demo part of the functionality of the 3D Tiles tools.


javagl:
The last commits change the former "PackageServer" demo into an actual command line functionality.

Running
`npx 3d-tiles-tools serve -i ./specs/data/packageServer/3tzTests`
will serve this directory, so that the tilesets can be loaded with the [test sandcastle](https://github.com/CesiumGS/3d-tiles-tools/blob/a2ef67afe5e90ef27d975431e64bbc1800f2c2a8/specs/data/packageServer/3tzTests/Sandcastle.js). Of course, in addition to the input directory, it can already receive a `host` and `port` (with `locahost` and `8003` being the defaults).

One question that still has to be resolved: Which additional parameters should that command receive? 

I could imagine that it may require parameters for CORS handling. Right now, the server is setting some [permissive CORS headers](https://github.com/CesiumGS/3d-tiles-tools/blob/a2ef67afe5e90ef27d975431e64bbc1800f2c2a8/src/tools/packageServer/PackageServer.ts#L133), and that may have to be configurable.

Another option that could be added is some sort of "log level". In general, we'll have to decide _what_ should be logged, and _how_, and under _which_ conditions.

Another question that is still open is the one about [Entry compression in 3D Tiles packages](https://github.com/CesiumGS/3d-tiles/issues/806). For example, 3TZ says that the entries can be compressed with Zstandard or 'deflate'. But it's not clear how that should be handled by the server. For example: When a ZSTD compressed content is found, should it just be sent to the client, or should it be _uncompressed_ and then _re-compressed_ with GZIP? Even JPEG files, for which this hardly makes sense....? Right now, GZIP compressed content causes [the `"Content-Encoding" = "gzip"` header](https://github.com/CesiumGS/3d-tiles-tools/blob/a2ef67afe5e90ef27d975431e64bbc1800f2c2a8/src/tools/packageServer/PackageServer.ts#L255) to be inserted, but given that https://github.com/CesiumGS/3d-tiles/issues/806 is not resolved yet, that's a pretty arbitrary "Why not? 🤷‍♂️"-approach.

---

An aside: I was actually a bit surprised that it managed to run the specs for this in CI. The specs are starting such a server on localhost, running some trival "client" to iterate over all explicit tiles of the tilesets, and collecting the URLs that could _not_ be resolved (which should be an empty array, of course).

javagl:
There are now two additional command line options:

- `cors` to enable CORS
- `developmentMode` to disable caching and print additional logging information

Also, the server did previously not properly handle the case that the client requested a `3tz` file _directly_ as in
`http://localhost:8003/tileset.3tz`
It only worked with
`http://localhost:8003/tileset.3tz/tileset.json`
The resolution mechanism is a bit tricky when `3tz` files are supposed to be served "directly": The URL replacement has to check whether a URI refers to _another_ 3TZ file, and only add the 3TZ file name prefix if this is _not_ the case. I hope that the extended tests cover this reasonably well: The sandcastle now includes a test that directly refers to a 3TZ file (even though _this_ is not tested for all possible cross-references among them).

lilleyse:
Thanks @javagl!

jo-chemla:
Nice to see the `serve` CLI now part of `3d-tiles-tools` to serve `3TZ or 3DTILES` packages. 

Just out of curiosity, are there any plans to add support directly within the [cesiumjs](https://github.com/CesiumGS/cesium) library / or broader [cesium-native](https://github.com/CesiumGS/cesium-native) ecosystem - unreal engine plugins etc? 

The 3tz already has a community specification at [Maxar-Public/3tz-specification](https://github.com/Maxar-Public/3tz-specification/blob/main/Specification.md), should `*.3dtiles` also be specified before they get into CesiumJS (see thread https://github.com/CesiumGS/3d-tiles/issues/727)? Maybe similar to how recently the quantized-mesh format has also been specified at [quantized-mesh/SPECIFICATION](https://github.com/CesiumGS/quantized-mesh/blob/main/SPECIFICATION.md). 

```sh
npx 3d-tiles-tools serve -i ./specs/data/packageServer/3tzTests
```

(somewhat related, quick mention of packages as external tilesets https://github.com/CesiumGS/3d-tiles/issues/804)

javagl:
@jo-chemla There are several interconnected questions here. Maybe we can sort out some of them, and either add more specific points to the linked issues, or open new ones.

About 3TZ:

The 3TZ specification has been published quite a while ago, and is indeed very stable. (And I'm saying that despite a recent update - this update mainly added `.3dtiles.zip` as a possible file extension, so there are no _structural_ changes associated with that). And there recently have been some efforts for a broader support of 3TZ in Cesium. For example, as mentioned in the [May 2026 release notes](https://cesium.com/blog/2026/05/01/cesium-releases-in-may-2026/#cesium-ion-and-tilers-updates), Cesium ion Self-Hosted can now ingest and stream 3TZ data (with that 'streaming' basically offering the same functionality as the 'package server' here). 

About 3DTILES:

The '3DTILES' (SQL-based) package format is also ... somewhat stable. It is already used in various places internally, and 'not going anywhere' in terms of its _structure_. And it's a bit unfortunate that this is already used, but not really publicly and reliably _specified_. The linked issue was an attempt to do so.  That specification PDF is a bit more restrictive than what is currently used, but this refers to a few technical details. I'd also really like to see this become **the** official specification, but ... most of that is simply not my call.

About "packages as external tilesets":

On the one hand, there are some subtleties involved here, but again, for the case of 3TZ, this is already pretty settled: There is the [`MAXAR_content_3tz`](https://github.com/Maxar-Public/3d-tiles/tree/MAXAR_content_3tz/extensions/MAXAR_content_3tz) extension, which is already used in the [tests of the 'package server' here](https://github.com/CesiumGS/3d-tiles-tools/blob/61e364761b19279df93d76fa3b3c036821a263e7/specs/data/packageServer/3tzTests/README.md)

The big one: Direct support in CesiumJS and cesium-native:

I'm curious what _exactly_ you mean by "direct support". 

In general, these files are "archives/packages" for complete tilesets. And of course, the examples here are just "dummy" examples, but in reality, such files often have many megabytes, sometimes several gigabytes, and files with several _terabytes_ are already on the radar. 

Clients like CesiumJS can not do anything with such a file. Even though it would be _possible_ to implement the functionality that is already implemented _here_ in CesiumJS (namely, of opening such a file and accessing its entries), this would be limited to ~"what a browser accepts as a 'file'", i.e. whatever it is, it can not be bigger than 2GB, and opening CesiumJS and seeing it download a 2GB file before the first LOD appears doesn't make sense. 

The goal of the 'package server' therefore was to set such a file "appear like" a standard, non-packaged tileset, to be accessed by a client, over the network.

One could make a case to add further support for such packages in `cesium-native`, though: Native applications in general **do** have access to 'files', and they _could_ directly read from a 3TZ, without the detour of having a local server running to deliver that data via network. Use-cases for this could be things like CesiumForUnreal applications that are bundled with one large 3TZ for all their data, and that they read directly and internally, by just double-clicking that `.exe` file. 

I think that some of that wouldn't even be such a heavy lift: The infrastrucuture behind 'cesium-native' can relatively easily be tweaked to access local file (a _very_ basic proof-of-concept [in `cesium-cpp`](https://github.com/javagl/cesium-cpp/blob/36ed45f0697a2f28351fdfd04f8ab2b5031a62d3/Source/src/CesiumNativeImpl/FileAssetAccessor.cpp)). But implementing this properly and adding actual 3TZ support would require some work, and I'm not aware of any active efforts in this direction.

jo-chemla:
Thanks as always for the detailed reply. 
Great to see that `3DTZ` is mature and structurally stable, with the maxar spec already existent/public and easy to refer-to, and that `3DTILES` is also structurally pretty stable, although more loosely, at least less publicly, specified. 

Regarding the direct integration within CesiumJS/cesium-native, I was indeed not very explicit. 
Nice idea for cesium-native being able to use the `3d-tiles-tools serve` middleware or equivalent on startup, to stream tiles from the package, to be consumed by the current library. Indeed not doable for the web where these hundreds of GBs or even TBs of package cannot be sent over network.

What I had in mind is rather a port of the concept of 2D [COG](https://cogeo.org/) (Cloud Optimized Geotiff), 3D [COPC](https://copc.io/) (Cloud Optimized PointCloud), or other [CNG](https://github.com/cloudnativegeo/cloud-optimized-geospatial-formats-guide) (Cloud Native Geospatial, usually referred to for encompassing vector, raster, 3d etc) formats. These are extensions built on top of older, already well standardized and mature formats (Geotiff for COG, LAS/LAZ for COPC etc - by the way just realized I did not get back on the other thread about COPC), that organize the data within the file, and expose file header to quickly access specific bits of data. My take on the background behind this: 
 - for COG, data is pyramided, akin to mipmapping, storing the original resolution image as well as down-sampled resolutions, usually a 2x resolution decrease at every level of the hierarchy
 - also important, data uses block tiling scheme rather than storing by rows or columns of pixels, so pixel area close to one-another are stored in contiguous memory blocks. 
 - the file header content specifies CRS, blocksize, pyramid level resolutions etc, so that a consumer client (for viewing or processing) can access the memory block(s) containing a subsampled resolution of the data, or a specific small AOI within the image at higher resolution etc. 
 - Historically, people were often relying on a middleware like [TiTiler](https://developmentseed.org/titiler/), to expose a standard TMS/XYZ WebMercator tiles endpoint from a COG, with the middleware doing all the heavy lifting - FastAPI python backbone accessing the input COG blocks required for creating the tile, at most appropriate resolution, concatenate, reproject, and clip to export XYZ tiles. 
 - The client side tooling is now mature enough to stream directly client-side these massive COGs (sometimes a few hundreds GBs) eg [QGIS](https://courses.spatialthoughts.com/qgis-cloud-native-geospatial.html#load-a-cloud-optimized-geotiff-cog) or even to the web via eg the great devseed folks like kylebarron at [deck.gl-raster](https://developmentseed.org/deck.gl-raster/blog/v0-7-release/), maplibre specific [geomatico/maplibre-cog-protocol](https://github.com/geomatico/maplibre-cog-protocol) or agnostic [geotiff.js](https://geotiffjs.github.io/)
 - On the web, these libs use [HTTP Range Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests) to target specific bits of the file to pass-on to the rendering layer very efficiently, close to viewport resolution to avoid unnecessary network transfers. The nicety is that like 3Dtiles directory of files hierarchy, and unlie other tiled image servers (WMTS, IIIF etc) only a minimal, pure file server is required, with range request support.
 - Same for COPC with desktop or web implementations for viewing/consumption eg [potree COPC](https://github.com/potree/potree/pull/1381) or [qgis](https://docs.qgis.org/3.44/en/docs/user_manual/processing_algs/qgis/pointcloudconversion.html), or [PDAL](https://pdal.org/en/stable/stages/readers.copc.html) for accelerated processing
 - A similar philosophy, and probably a lot closer to what `**.3tz/*.3dtiles` direct integration could be, can be found in massive image formats viewers, for example in the [Single-file Tile formats](https://openseadragon.github.io/examples/creating-zooming-images/) section of OpenSeadragon. The `SZI` is indeed a simple `losslessly zipped DZI container`, where DZI is the image tile files directory. So in theory, using similar concepts/implementations, the cesium-js or cesium-native libraries could access the specific portion of the zip - easier if stored with `compress=none`, but also doable with some compression according to some discussions in this long [osd thread](https://github.com/openseadragon/openseadragon/issues/944). 
 - Note the multiple benefits: if client applications do not understand the COG/COPC specs but still talk Geotiff or LAS, they can still open the file without benefitting the structural acceleration. Moving massive file counts directories is also often a pain, with massive overhead and very slow transfer speeds of 100.000s of files or the need to pack/transfer/unpack - which was a core motivation for zoomable images single-flie archives. Often result in simplifed dataset sharing and access. And faster processing/use-cases for extracting by-products of the data, most often downsampled/substracts of the data. 


Now realizing this is a lot more work than anticipated, but the needs are shared with other communities (geospatial 2d, microbiology 2D, geospatial 3d etc), and this would definitely deserve an issue thread of its own! Not sure if it should be within [cesiumGS/cesium](https://github.com/cesiumGS/cesium/) (I would say so), or other repos 3d-tiles-tools, native, I'll probably duplicate the comment there!

javagl:
There's a **lot** to unpack in this comment. 

I'm not sure how deeply I could or should zoom into each of the references in order to full grasp the extent of the topic. I had a quick look at most of the links. But even just [one of them](https://openseadragon.github.io/examples/creating-zooming-images/) consists of a long list of tools for different formats and tasks in different languages, and randomly clicking on one of them leads to a repo with the last commit being 10 years ago...
This is the _implementation_ side. 
The different _specifications_ are even harder to fully understand, in terms of 1. how to align that with package format specifications and 2. (after that) possible support for any of that in the 3D Tiles Tools.

Trying to look at this from a slightly higher abstraction level:

To my understanding, the most important difference between the listed formats and the current (3TZ/3DTILES) package formats is that the current package formats are pure _containers_, without any semantics for the contents. The other formats are built around the (hierachical) _structure_ of the data that they contain.

Indeed, 3TZ is a plain ZIP. The only difference is that it contains the "index file" that allows random access to individual entries of the ZIP, based on their names. 3DTILES is a plain 2-column-database. This also aims at providing that plain `fileName->blob` lookup. Neither of them are specific for 3D Tiles in any way.... **except** for that they must contain a `tileset.json`. (Which, when looking at it like that, is an oddly specific and nearly "unnecessary" or "artificial" requirement...).

And **iff** I understood this correctly, then an important point of the "client-side support" could be that the information about the structure (and the layout of the data within the file) could be transported to the client in a form that would allow the client to do _range requests_. 
(This is not applicable for 3DTILES. It has been considered for the "3TZ-based" server. But given that the 'interface' of the 3TZ file format is only that of the `fileName->blob` lookup, range requests are not _directly_ applicable there either)

The questions about how clients could "support" (which of) these formats, and what that would entail in terms of the implementation effort for the clients and servers is going beyond what I can quickly address here. But I think that one benefit of the current package formats is exactly that the client can remain "dumb". It's requesting a `tileset.json`. It's requesting a `content.glb`. It does not know anything about package formats. And I think that this is a **very** important benefit. Nobody wants to pull in 10 libraries and special handling paths for different file formats into their client... _unless_ this special handling offers a benefit that is worth this effort.

More specifically: My gut feeling is that there are still some (unspoken) assumptions about usage patterns. For example...

>  if client applications do not understand the COG/COPC specs but still talk Geotiff or LAS, they can still open the file without benefitting the structural acceleration. 

I have the impression that this is somewhat revolving around 'Desktop applications'. (CesiumJS does not know what a 'file' is...). Maybe that could be one dimension along which this (complex) topic could be broken down into smaller pieces. The current package formats clearly revolve around the assumption that the only 'interface' between the client and the 'file' (server) is that the client sends out the 'name' of the file that it needs, and receives a 'blob'. This does not mean that the client could not have a "deeper" understanding of the format. For example, instead of sending out a 'name', it could send out a range request to access a chunk of the file (as a blob) more directly, if it "knew" the structure of that file. 

Other aspects of package formats are independent of all that - for example, the following is uncontroversial:

> Moving massive file counts directories is also often a pain, with massive overhead and very slow transfer speeds of 100.000s of files or the need to pack/transfer/unpack -

And this is _strongly_ related to https://github.com/CesiumGS/3d-tiles/issues/598 : When using implicit tilesets, it can easily happen that there are "waaay too many" (t/f)iles in one directory. And unfortunately, the current specification and implementation do not offer a sensible way to avoid that: Even when anticipating the exponential growth, and using the most fine-grained directory structure that is possible, the "leaf layer" may easily exceed the number of files that can be stored in one directory. (And this is a technical constraint - i.e. this is on top of the fact that _copying_ millions of files is awfully slow)

---

> Now realizing this is a lot more work than anticipated [...] and this would definitely deserve an issue thread of its own!

I agree. Given the complexity (existing specifications, usage patterns (Desktop/web), possible alignment with the current package formats, impliciations for clients...), it's hard to say what's the best place for all this. But maybe a first, high-level issue could go into the `3d-tiles` repo, just to sort things out...?

jo-chemla:
Probably the discussion can actually be moved over to https://github.com/CesiumGS/3d-tiles/issues/727 which already did exist. Although this 727 thread might be deemed close-able as the `3tz/3dtiles` formats do exist, and the new topic is rather about consuming these formats without the `3d-tiles-tools serve` middleware into user-facing apps (web/native), which could justify a new issue thread being opened in `3d-tiles` repo. 

I might have put too much context in the previous posts - in which case sorry about that - mostly to showcase that some Open Geospatial Consortium specified open-standards do exist (COG, COPC), made to store, stream and consume single-file packaged, massive 2D/3D datasets, from web browser or native apps. 

The references to OpenSeaDragon were mostly to highlight shortcuts used by the deepzoom communities for streaming single-file massive tiles collections, via mostly the SZI format - barely an uncompressed zip of a DZI directory structure, which can be queried for directory structure as well as raw file contents within the whole file efficiently. 

You're right about everything from my pov: 
 - the distinction between `containers/structural` formats stands. SZI is a pure container (uncompressed zip), COG/COPC are indeed internally tiled - more akin to a 3Dtiles implicit tiling, COPC using octree, COG using quadtree resolution pyramid. 
 - true that the metadata information about the structure has to be queriable by the client before range requests HTTP2 can be made smartly. COG/COPC allocate a memory block within the file, that is queried first so the underlying data structure is known and the right portions of the file / right tiles can be queried depending on the user viewport/screenSpaceError. SZI uses standard zip libraries to retrieve this directory structure
 - the fact that the client can remain dumb is what's behind the need for [TiTiler](https://titiler.xyz/), the COG middleware that consumes a COG remote file, and serves a standard TMS/XYZ endpoint. So apps that can only talk TMS can point to titiler endpoint, and retrieve tiles encoded from the cog on the middleware side. There has however been a ongoing shift for Cloud Native Geospatial standards to be consumed/queried directly within the GIS softwares or web libraries, by implementing these specs client-side. These were the links to [deck.gl-raster](https://developmentseed.org/deck.gl-raster/blog/v0-7-release/), [geotiffjs](https://geotiffjs.github.io/). STAC catalogs of the whole ESA/NASA Landsat/Sentinel products, and a lot other earth-data roducts, are served as COGs directly, with no middleware, and QGIS, maplibre etc can consume all items as-is. Same for USGS 3DEP Lidar programme data served as COPC, consumed directly by qgis/tiled 3D libraries. 
 - True that the quote is about desktop apps, ingesting massive COGs/COPCs without benefitting the underlying acceleration structure. Here the subject of the feature requested is rather focused on ingesting `3tz/3dtiles` packages without the `3d-tiles-tools serve` middleware. 
 - https://github.com/CesiumGS/3d-tiles/issues/598 is somewhat related but, as I understand it, just to limit the count of tiles/files **per-folder**. The total count of files of the tileset would probably remain pretty high, in the hundreds of thousands depending on the tileset (say 10B triangles 3d-scan max 50k tris per tile, the leave tiles files only would be roughly 200k, plus the intermediate levels). Even if it helps to store these file counts dispatched over several folders, the total file count will still be intensive for file operations (copy/move to cloud-storage etc). 

The Cesium-unreal issue could probably be opened on the `cesium-unreal` repo, now that your 3D-tiles-tools middleware exists, and could probably be simpler to implement/wrap in case the user wants to stream from a `3tz/3dtiles` package rather than `tileset.json`. 

For illustration, here is the memory blocks of a COPC file, with metadata about the structure at the end of the file.

javagl:
Yes, given that this is "only" a small PR for essentially moving something from the 'demos' to the 'src' folder, broader discussions might easily be lost or not properly be "channeled" into some "open->discuss/address->resolve" lifecycle of an actual issue. Maybe the core points (or more specific suggestions) can be moved into dedicated issues as appropriate.

But of course, there is a connection to the 'tools' on the level of the _implementation_. For 3TZ, I think that the only (?) publicly available _implementation_ is in the [`ArchiveFunctions3tz.ts`](https://github.com/CesiumGS/3d-tiles-tools/blob/37f69bb61bb3cd214a85d068285154296d7f3414/src/tilesets/packages/ArchiveFunctions3tz.ts) here.  

(This code was only "ported" from an old implementation that was made to the (then: ) 3d-tiles-validator, and this was done very pragmatically, without any refactorings or cleanups. Afterwards, some bugfixes went into that, but the implementation still has a fundamental flaw insofar that it does not handle hash collisions. I hope that people will **not** consider this as a "reference implementation", because that might proliferate the hash collisions issue. There is a somewhat "cleaner" implementation of all that in a (Java) project that is called 'j3dtiles', but I'm keeping that 'private' for now)

One (maybe important) note: There are currently several interconnected issues about glTF 2.1 and 3D Tiles 2.0, with the latter roughly "being a set of glTF extensions". An the specification of glTF 2.1 will include a new binary format definition. This will (also) not directly aim at _most_ of the topics that have been mentioned here (streaming, random access, etc), but one has to expect that any new 3D Tiles "package format" will have to be aligned with that in one way or another.

(Related issues are https://github.com/CesiumGS/3d-tiles/issues/822 and https://github.com/KhronosGroup/glTF/issues/2585 )

PR #90, opened 2024-01-19
Converting the monorepo back to CommonJS

javagl:
The PR https://github.com/CesiumGS/3d-tiles-tools/pull/64 converted this repository into a monorepo containing ESM modules. 

Unfortunately, it is not reasonably possible to use ESM modules from CommonJS projects. In order to "emulate" a trivial import and usage like
``` 
import { TilesetSources } from "@3d-tiles-tools/tilesets";

function example() {
    const source = TilesetSources.createAndOpen("tileset.3dtiles");
    ...
}
```
it is necessary to come up with contraptions like this....
```
const tilesetsImportPromise = eval("import('@3d-tiles-tools/tilesets')");
const tilesetSourcesImport = async () => {
  const tilesetsImport: typeof import("@3d-tiles-tools/tilesets") =
    await tilesetsImportPromise;
  return tilesetsImport.TilesetSources;
};
async function example() {
  const TilesetSources = await tilesetSourcesImport();
  const source = TilesetSources.createAndOpen("tileset.3dtiles");
  ...
}
```
which is plainly unusable (and has a bunch of additional limitations and issues, but I think that it is not even worth to go into the details here). While there are claims that it is possible to compile a project to CommonJS and ESM simultaneously, they usually involve nesting several layers of build- and transpilation systems that each come with their own quirks.

When changing the type from `module` to `commonjs` (and removing the `.js` file extensions in each and every import), everything seems to _just work_. So this looks like the more reasonable trade-off for me right now.


javagl:
I have tested the resulting packages as dependencies...

- in a TypeScript project with `type: module`
- in a TypeScript project with `type: commonjs`
- in a JavaScript project with `type: module`
- in a JavaScript project with `type: commonjs`

and they seem to work smoothly in all of them. For me, right now, there does not seem to be a strong, technical reason to insist of ESM. 

(An aside: after changing it to CommonJS, the workaround from https://github.com/CesiumGS/3d-tiles-tools/pull/64#issuecomment-1854267607 is no longer necessary)

I'll clean up the documentation and README in _this_ branch accordingly.

(One thing that I still wanted to try out: Whether it is possible to omit the `@3d-tiles-tools/cli` package, and let this package just **BE** the `3d-tiles-tools`)

Other than that, I think that I'll merge this one into the main PR soon.

PR #96, opened 2024-01-31
Fix metadata migration of null strings

javagl:
(Builds upon https://github.com/CesiumGS/3d-tiles-tools/pull/64 )

When the batch table of a B3DM contains `null`/`undefined` strings, then the migration tried to convert them, and in that process, called `encode` on these strings, causing all byte offsets to be shifted by the length of the string `"null"`.

This fix only causes `null`- and `undefined` strings to be "skipped" when computing the offsets.

This does not (yet) include spec data. It might be necessary to create a B3DM with a batch table that contains exactly this type of data, with `null` values. If desired, I can allocate some time to create one.


javagl:
The respective line was taken from https://github.com/CesiumGS/cesium/blob/edf71462a2f0761de140d786dfebb880612f5cf8/Specs/MetadataTester.js#L483 , so... maybe that should also be fixed and/or the test coverage be increased there, to include the case of "null" strings...

PR #101, opened 2024-02-17
Undo monorepo conversion

javagl:
This is essentially one big **"CTRL+Z"** for the things that have been done in https://github.com/CesiumGS/3d-tiles-tools/pull/64 . The goal is to eventually have a single `3d-tiles-tools` package, because there have been some open questions about establishing an organization called `3d-tiles-tools` on NPM for publishing the scoped `@3d-tiles-tools/...` packages.

For now, the contents of the workspaces/packages are moved back into a "flat" `src` folder. The build infrastructure (i.e. the individual `tsconfig.json` and `package.json` files) is removed. Further details will be added here soon.


javagl:
The conversion back to a "not-monorepo" is done.

The original workspaces/packages have been moved back into the top-level `/src` folder. 

---

Some details:

The directory structure _within_ this folder _might_ look a bit unusual at the first glance: There are files like
`src/base/base/Buffers.ts` or
`src/ktx/ktx/KtxUtility.ts` 
The reason for that is that the _first_ subdirectory corresponds to the "package name", and the _second_ subdirectory level reflects the structure _within_ that package. Each of the _first_ subdirectory levels contains an `index.ts` file, like
`src/base/index.ts`
that exports the "API". One reasoning behind that is that _cross-package imports_ can be written in the form
`import { BinaryBufferStructure } from "../../base";`
which at least more closely resembles the
`import { BinaryBufferStructure } from "@3d-tiles-tools/base";`
from the packaged version. The alternative would be to descend into the respective _subdirectory_ _within_ the "package", as in
`import { BinaryBufferStructure } from "../../base/binary/BinaryBufferStructure";
Whether or not this _really_ makes sense or is a good practice? I'm open for comments here...

---

I tested running the `npm install` and `npm run package` scripts (multiple times...). 
I've tried out using the resulting package as...
- a basis for installing an running the tools as a command line application
- a dependency in a `CommonJS` JavaScript project
- a dependency in a `ESM` JavaScript project
- a dependency in a `CommonJS` TypeScript project
- a dependency in a `ESM` TypeScript project
- a dependency for the `3d-tiles-validator`.... which actually required **zero** code changes. 

I also ran all `demos`, and did a short review/cleanup pass of the README and IMPLEMENTATION, and have been ballsy enough to already put a release date into the `CHANGES.md`. Let's see how this goes...

Issue #134, opened 2024-05-28
Consider a command to fix certain alignment errors


javagl:
This is a spin-off from https://github.com/CesiumGS/3d-tiles-tools/issues/5 : The requirements for [Padding](https://github.com/CesiumGS/3d-tiles/issues/225) in 3D Tiles - specifically, certain alignment requirements for tile content files like B3DM - have made their way into the specification relatively late. And there are still _many_ B3DM files out there that do not meet these requirements. These files will be declared to be 'invalid' by the 3D Tiles Validator.  

(In fact, even the spec files for CesiumJS did not meet these alignment requirements, and have been [fixed](https://github.com/CesiumGS/cesium/commit/4b3b4186b4afa2307d7d1666569826d864b984ca) only recently...)

While CesiumJS can process these files even when they do _not_ meet these requirements, there is a huge potential for inconsistencies and fragmentation: Implementors should **not** be "forced" to be able to read "invalid" files.

Many of these alignment issues could be fixed relatively easily: The 3D Tiles Tools already offer the functionality to 

- [read tile data](https://github.com/CesiumGS/3d-tiles-tools/blob/9ecafcf20821113898ce88f78160be8e02a70375/src/tilesets/tileFormats/TileFormats.ts#L114) from buffers containing B3DM/PNTS/I3DM files into a [`TileData`](https://github.com/CesiumGS/3d-tiles-tools/blob/9ecafcf20821113898ce88f78160be8e02a70375/src/tilesets/tileFormats/TileData.ts) object
- [create a buffer](https://github.com/CesiumGS/3d-tiles-tools/blob/9ecafcf20821113898ce88f78160be8e02a70375/src/tilesets/tileFormats/TileFormats.ts#L510) containing B3DM/PNTS/I3DM data from a `TileData` object

The latter will already ensure most (all?) of the common alignment requirements.  

So fixing the alignment could be a one-liner...
`fs.writeFileSync("out", createTileDataBuffer(readTileData(fs.readFileSync("in")))`
which should unfold only into "a few" _real_ lines of code, with input/output type checks, 'overwrite' checks, and a proper command line command.

javagl:
Fixed via https://github.com/CesiumGS/3d-tiles-tools/pull/136

PR #136, opened 2024-06-01
Add function to update alignment

javagl:
Addresses https://github.com/CesiumGS/3d-tiles-tools/issues/134 : 

This adds a command `updateAlignment` that can be applied to B3DM, I3DM, PNTS, and CMPT files, and just ensures that the elignment requirements for the batch- and feature table and the tile data as a whole are met.

I tested it with an old version of the CesiumJS spec data (the one before [this commit](https://github.com/CesiumGS/cesium/commit/4b3b4186b4afa2307d7d1666569826d864b984ca)), and it fixed most of the validation errors from that state. (Adding a spec could make sense, but ... creating the "invalid" data would take most of the time here...). In any case, the README.md still has to be updated.


javagl:
I added specs based on the "old" CesiumJS specs data (one B3DM, I3DM, PNTS, and CMPT). They all caused validation errors before. The specs are comparing the result of running `updateAlignment` to the versions that are contained in the `golden` subdirectory, which no longer cause validation errors.

The README.md has been updated to include the `updateAlignment` command.

(An aside: One could make a case for adding this "alignment update" as part of the `upgrade` command. Technically, this would be trivial. But some details might have to be sorted out in terms of the desired configurability (even though the update options are not exposed right now beyond the `targetVersion` anyhow). Cases where someone wants to fix the alignment of a B3DM _without_ upgrading its glTF 1.0 to 2.0, or someone wants to fix the alignment for B3DM and not for PNTS are unlikely, so it might just be part of the detault `targetVersion=1.0` update...)

Issue #138, opened 2024-06-19
Invalid path handling for `combine` command


javagl:
This is a follow-up of https://github.com/CesiumGS/3d-tiles-tools/issues/137 , with the actual issue being isolated and carved out into a test case. 

There are configurations where the `combine` command generates invalid paths/URIs for the tile content. The (distressingly simple) example to reproduce the error is here:

[externalInSameDirectory-2024-06-19.zip](https://github.com/user-attachments/files/15904144/externalInSameDirectory-2024-06-19.zip)

The structure is
```
tileset.json refers to
  sub0/externalA.json
  
sub0/externalA.json refers to
  tileA.b3dm
  externalB.json
  
sub0/externalB.json refers to
  tileB.b3dm
```

The `combine` command has to combine external tilesets _recursively_. This means that it will...

- traverse the root `tileset.json`
  - traverse the `externalA.json`
    - update the URI for `tileA.b3dm` to `sub0/tileA.b3dm`
    - traverse the `externalB.json`
      - update the URI for `tileB.b3dm` to `sub0/tileB.b3dm`
    - finish traversing `externalB.json` and put its root into `externalA.json`
  - **continue** traversing `externalA.json` ...
    - update the URI for `sub0/tileB.b3dm` to `sub0/sub0/tileB.b3dm` ⚡ 
 
If I got this right, then the issue is related to the fact that the root of the processed external tileset is put into the surrounding tileset, and therefore traversed _again_, updating the URI for the tile content _twice_. 

There is no trivial or obvious way of detecting whether a tile structure that is encountered during the traversal already had been there from the beginning, or whether it occurred due to "inlining" another external tileset - so it is not clear which content URIs have to be updated. 

However, I think that there is a trivial way of solving this.

(I have no idea _how_. "Obvious" approaches of ~"not traversing the inlined parts" will break in other configurations. But I'm pretty sure that there _is_ a trivial way, and I'll try to find it...)

javagl:
There may be other issues with the `combine` command, related to _external tilesets_ and _multiple contents_. 

[...]

Issue #43, opened 2023-06-10
[Feature Request] `TilesetCombiner` fails for older tilesets with content adressed by `url` property instead of `uri`


jo-chemla:
When using the `combine` method on tilesets produced by ContextCapture, then an error `Content does not have a URI` is raised. This is probably because the tile contents in the tileset root json file (plus subsequent tiles) are referred to via the `url` property rather than the `uri` property - which references either b3dm or json tile contents. 

I know I can use the 3d-tiles-tools `upgrade` utility on local tilesets I do manage - although it will take a lot of time to upgrade loads of tilesets. 
But for tilesets I do not manage, which are stored online on different servers and that I'd like to combine, then the easiest way would be for the `combine` utility to allow for tilesets which use either the `url` or the `uri` property to reference tile content.

javagl:
We have been pondering this question internally. 

The pragmatic view is: All inputs have to be valid. And for the case of the `uri` vs. `url`, the recommendation would be: Run `upgrade` on the input, and then use the upgraded tilesets for any further operations. 

**But...** I can see that this is... invonvenient, particularly when the reason for something not working properly is literally that of a single letter (or "a pixel", as in "the difference between i and l...). 

One reason of why we are (currently) not just ignoring this point is that the generalization of this question could be: Which forms of "invalid" inputs should be "ignored"? Or: Which parts of the `upgrade` should be done autmatically in other operations? At some point, we'd have to spread workarounds for legacy data, handling for special cases, and pseudo-upgrade functionality throughout the code. And it would _really_ be better to summarize this in the `upgrade` command. 

(One example: When you `combine` two tilesets that use `url` - should the combined tileset use `uri` or `url` then? It should probably use `uri`, but maybe you _want_ to retain the original `url`...?)

[...]

Issue #137, opened 2024-06-13
i have error about combine command.


JHP0124:
![2024-06-13 11 50 36](https://github.com/CesiumGS/3d-tiles-tools/assets/105851614/188f67bb-f240-44f5-b391-85febaf504e2)

![image](https://github.com/CesiumGS/3d-tiles-tools/assets/105851614/52646d72-9e1c-47ae-9893-7906fc6dea6d)

![image](https://github.com/CesiumGS/3d-tiles-tools/assets/105851614/09e89114-1989-4e8e-9f8c-bad70c566565)

this is example structure picture .

and i attached json file and please check json file
[tileset.json](https://github.com/user-attachments/files/15813140/tileset.json)
[Tile_+000_+021.json](https://github.com/user-attachments/files/15813145/Tile_%2B000_%2B021.json)

so i tried to combine this dataset.

but... 3d-tiles-tools show me this error.

![image](https://github.com/CesiumGS/3d-tiles-tools/assets/105851614/66b2bcab-77d3-4aad-908c-ec8ed9ea6f3e)

:~/light$ npx 3d-tiles-tools combine -i ./ -o ./combine_test
Need to install the following packages:
3d-tiles-tools@0.4.2
Ok to proceed? (y) y

npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
[11:47:51.808] ERROR (CLI): No data found for Tile_+001_+021/Tile_+001_+021/Tile_+001_+021_L20_0uuuu20.b3dm
    err: {
      "type": "TilesetError",
      "message": "No data found for Tile_+001_+021/Tile_+001_+021/Tile_+001_+021_L20_0uuuu20.b3dm",
      "stack":
          Error: No data found for Tile_+001_+021/Tile_+001_+021/Tile_+001_+021_L20_0uuuu20.b3dm
              at TilesetCombiner.combineContentInternal (/home/anjion/.npm/_npx/e97e2e75cde22e22/node_modules/3d-tiles-tools/build/src/tools/tilesetProcessing/TilesetCombiner.js:168:19)
[...]

PR #139, opened 2024-06-22
Fix combine with multiple contents

javagl:
Fixes https://github.com/CesiumGS/3d-tiles-tools/issues/138

The issue originally referred to an aspect of the URI handling that was not specific for multiple contents, but has been extended to cover some related shortcomings of the current implementation of the `combine` command.

Fixes https://github.com/CesiumGS/3d-tiles-tools/issues/43

We talked about whether the (legacy) `url` property should be accepted in general, or whether updating this to `uri` should be done with a dedicated `upgrade` pass. But the `combine` command has to twiddle with (and modify) the `uri` properties of contents anyhow, so I think that this is a case where _accepting_ the legacy form (and printing a warning) is OK...

---

The original implementation did not properly update the URIs of contents that had been found in external tilesets. This was caused by the fact that the 'combine' command can add new children to a tile, _while_ the tile is being traversed. These new children had already been traversed during the recursive combination of nested external tilesets, causing some tiles to be traversed _twice_.

(There are some potentially deep, subtle intricacies when it comes to _structural modifications of a tile hierarchy *during* the traversal_: What should be allowed, and what should the behavior be? Originally, the implementation used the `Tiles.traverseExplicit` utility funciton, which does not anticipate structural changes. Now, the traversal has just been moved into the `TilesetCombiner`, and is implemented there in a way that handles the modifications that it does)

---

A broader issue with the `combine` command was that of handling _multiple contents_. Until now, when an external tileset was found, then the properties (`content`/`contents`) of the tile _containing_ this external tileset had been replaced with the properties of the root of the external one. This breaks when there is a tile with contents `[tile.b3dm, external.json]` (which is obvious... in hindsight...). 

Now, the command properly handles this case, and the specs have been extended with a test case for this (as explained in the README of that spec data): 

- A tile refers to a SINGLE content, which is an external tileset
  - In this case, the properties of the tile will be "replaced" with the properties of the external root
- A tile refers to an multiple contents, including external tilesets
  - In this case, the external roots will be added as children to the tile

### DIFF
--- a/specs/data/combineTilesets/externalInSameDirectory/README.md
+++ b/specs/data/combineTilesets/externalInSameDirectory/README.md
@@ -1,0 +1,18 @@
+A regression test for https://github.com/CesiumGS/3d-tiles-tools/issues/138 
+
+```
+tileset.json refers to
+  sub0/externalA.json
+  
+sub0/externalA.json refers to
+  tileA.b3dm
+  externalB.json
+  
+sub0/externalB.json refers to
+  tileB.b3dm
+```  
+
+The content URI for `tileB.b3dm` of the combined result should be `sub0/tileB.b3dm`
+
+  
+  
--- a/specs/data/combineTilesets/externalInSameDirectory/sub0/externalA.json
+++ b/specs/data/combineTilesets/externalInSameDirectory/sub0/externalA.json
@@ -1,0 +1,32 @@
+{
+  "asset": {
+    "version": "1.1"
+  },
+  "geometricError": 4.0,
+  "root": {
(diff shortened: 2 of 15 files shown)


javagl:
I tried the new functionality with the tileset that was provided in https://github.com/CesiumGS/3d-tiles-tools/issues/137 and originally sparked this issue. It _seems_ to work. And I'll probably try it out by running the resulting tileset in CesiumJS and the validator. But ... now now. The resulting `tileset.json` has 200MB (!), which is... impractical in many ways...

Issue #44, opened 2023-06-10
[Feature Request] allow for `merge` Tileset utility to only produce root tileset json and avoid tile contents copy


jo-chemla:
Using the `merge` utility as presented in the docs starts by producing an output `tileset.json` root file which references the input tilesets in children directories. Then it continues by copying all the tile contents to the output folder. 

```bash
npx 3d-tiles-tools merge -i tileset_1/n_0.json -i tileset_2/n_0.json -o ./_merged_tileset
```

Since there are lots of files to copy, this copy operation can take a lot of time. Having an option to `disable tiles copying` (json + b3dm) would be great! Therefore, the output tileset would only reference the input tilesets - probably via a `uri` relative to the output tileset file. eg in the case above, the output tileset would look like 
```json
{
  "asset": {"version": "1.1"},
  "geometricError": 1,
  "root": {
    "boundingVolume": {...},
    "refine": "ADD",
    "geometricError": 1,
    "children": [
      {
        "boundingVolume": {...},
        "content": {
         // Important line: relative reference to merged tilesets
          "uri": "../tileset_1/n_0.json"
        },
        "geometricError": 1,
        "refine": "REPLACE"
      },
     // ...
    ]
  }
}
```

Note: this might be what the `combine` utility is made for, but since it does not process in my case because of a `Content does not have a URI` error (properties are `url`), I could not check, see this feature request #43

javagl:
This certainly could make sense for some cases. We also noticed that people occasionally want to "merge" tilesets, but expect the result to be what could be achieved by first calling `merge` and then (additionally!) calling `combine`. 

(And this should yield the right result in your case, if the input tilesets are "upgraded" to use `uri` instead of `url`, or we change the code to be able to handle this difference transparently - details in the linked issue).

There are some corner cases to consider here: 

Right now, most of the functionality of the `3d-tiles-tools` is targeted at being able to _transparently_ operate on the file system _or_ on 3D Tiles package files (i.e. `.3dtiles` or `.3tz` files). Very roughly:  You could run

`merge -i tilesetA.3dtiles -i tilesetA.3dtiles -o ./merged.3dtiles`

and receive a `.3dtiles` file that contains all the tiles from the input data (and it has to contain them to be "complete"). 

But there probably are some possible solutions for that:

[...]

PR #140, opened 2024-07-04
Add `jsonOnly` option to `merge` command

jo-chemla:
Following discussion thread https://github.com/CesiumGS/3d-tiles-tools/issues/44 

This option is meant to avoid copying resources (tileset.json and tile contents) when merging multiple input tilesets. Results in a tileset.json that references input tilesets where they are stored.

Can add another commit to remove the `--jsonOnly` flag from the `merge` command to instead create a new command, `mergeJson `

### DIFF
--- a/src/cli/ToolsMain.ts
+++ b/src/cli/ToolsMain.ts
@@ -553,14 +553,15 @@ export class ToolsMain {
     logger.debug(`Executing upgrade DONE`);
   }
 
-  static async merge(inputs: string[], output: string, force: boolean) {
+  static async merge(inputs: string[], output: string, force: boolean, jsonOnly: boolean) {
     logger.debug(`Executing merge`);
     logger.debug(`  inputs: ${inputs}`);
     logger.debug(`  output: ${output}`);
     logger.debug(`  force: ${force}`);
+    logger.debug(`  jsonOnly: ${jsonOnly}`);
 
     ToolsMain.ensureCanWrite(output, force);
-    await TilesetOperations.merge(inputs, output, force);
+    await TilesetOperations.merge(inputs, output, force, jsonOnly);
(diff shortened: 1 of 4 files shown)


javagl:
Thanks. I tried it out locally, and ran the tests, and verified that the CLA was signed, and everything seems to be fine....

...except for that line endings thing. Quickly looking at the [documentation](https://prettier.io/docs/en/options.html#end-of-line), it seems like `--end-of-line lf` is the default, so that should not be given explicitly. If this is removed, I can merge this PR. 

There's still the question about the line endings, though: I'll have to check what the best practices/policies are here. This should be unrelated to this PR, and may be handled in a small cleanup PR later.

jo-chemla:
Woops sorry for pushing that change in package.json. Indeed this was just testing on my side, I reverted to the original package.json prettier command.

jo-chemla:
Note I also tried enforcing line endings to `LF` in `.gitattributes` at the root directry of the repo, it might be a good option to have it in the repo for future external windows-based contributors (eol can also be set to auto). 

```bash
# .gitattributes
# Auto detect text files and perform LF normalization
* text eol=lf

```

javagl:
Yes, it seems like using a `.gitattributes` seems to be the way to go here. The desired behavior can also be configured by each user _locally_, but for the case that a user did not do this, the `.gitattributes` should take care of it. I'll confirm this internally, and likely add create a PR to add this soon.

Thanks again for the `mergeJson` functionality. 

(I'm usually updating things like the CHANGES.md and README.md prior to a release. So I'll probably schedule a release soon, and then also update these files accordingly)

jo-chemla:
Thanks for helping me design that PR, and happy that I could help!

PR #142, opened 2024-07-09
mergeJson: Set merged tileset children uri to relative path to output directory

jo-chemla:
In the initial PR, I did not correctly set the children URI for the merged tileset. This PR makes it so that children URI is the relative path from the output directory to the respective input tilesets jsons filepaths.

### DIFF
--- a/specs/tools/tilesetProcessing/TilesetMergerSpec.ts
+++ b/specs/tools/tilesetProcessing/TilesetMergerSpec.ts
@@ -90,8 +90,8 @@ describe("TilesetMerger", function () {
     actualContentUris.sort();
 
     const expectedContentUris = [
-      "specs/data/mergeTilesets/basicMerge/TilesetA/tileset.json",
-      "specs/data/mergeTilesets/basicMerge/sub/TilesetA/tileset.json",
+      "../../mergeTilesets/basicMerge/TilesetA/tileset.json",
+      "../../mergeTilesets/basicMerge/sub/TilesetA/tileset.json",
     ];
     expect(actualContentUris).toEqual(expectedContentUris);
   });
--- a/src/tools/tilesetProcessing/TilesetMerger.ts
+++ b/src/tools/tilesetProcessing/TilesetMerger.ts
@@ -142,12 +142,17 @@ export class TilesetMerger {
       }
       const tilesetSourceDirectoryName = path.basename(tilesetSourceName);
 
-      const tilesetTargetDirectoryPath = Paths.isDirectory(tilesetTargetName) ? tilesetTargetName : path.dirname(tilesetTargetName)
+      const tilesetTargetDirectoryPath = Paths.isDirectory(tilesetTargetName)
+        ? tilesetTargetName
+        : path.dirname(tilesetTargetName);
       const tilesetSourceIdentifier = TilesetMerger.createIdentifier(
         tilesetSourceDirectoryName,
         this.tilesetSourceIdentifiers
       );
-      const relativeTilesetIdentifier = path.relative(tilesetTargetDirectoryPath, tilesetSourceDirectoryPath);
+      const relativeTilesetIdentifier = path.relative(
+        tilesetTargetDirectoryPath,
+        tilesetSourceDirectoryPath
+      );
 
       const tilesetSource = TilesetSources.createAndOpen(tilesetSourceName);
       this.tilesetSources.push(tilesetSource);
--- a/src/tools/tilesetProcessing/TilesetMerger.ts
+++ b/src/tools/tilesetProcessing/TilesetMerger.ts
@@ -134,24 +134,26 @@ export class TilesetMerger {
 
       // Determine an "identifier" for the tileset source
       // (see `tilesetSourceIdentifiers` for details)
-      let tilesetSourceDirectoryName;
+      let tilesetSourceDirectoryPath;
       if (Paths.isDirectory(tilesetSourceName)) {
-        tilesetSourceDirectoryName = path.basename(tilesetSourceName);
+        tilesetSourceDirectoryPath = tilesetSourceName;
       } else {
-        tilesetSourceDirectoryName = path.basename(
-          path.dirname(tilesetSourceName)
-        );
+        tilesetSourceDirectoryPath = path.dirname(tilesetSourceName);
       }
+      const tilesetSourceDirectoryName = path.basename(tilesetSourceName);
+
+      const tilesetTargetDirectoryPath = Paths.isDirectory(tilesetTargetName) ? tilesetTargetName : path.dirname(tilesetTargetName)
       const tilesetSourceIdentifier = TilesetMerger.createIdentifier(
         tilesetSourceDirectoryName,
         this.tilesetSourceIdentifiers
       );
+      const relativeTilesetIdentifier = path.relative(tilesetTargetDirectoryPath, tilesetSourceDirectoryPath);
 
       const tilesetSource = TilesetSources.createAndOpen(tilesetSourceName);
       this.tilesetSources.push(tilesetSource);
       this.tilesetSourceJsonFileNames.push(tilesetSourceJsonFileName);
       this.tilesetSourceIdentifiers.push(
-        !jsonOnly ? tilesetSourceIdentifier : tilesetSourceName
+        !jsonOnly ? tilesetSourceIdentifier : relativeTilesetIdentifier
       );
     }
 


javagl:
I think that this was not really _changed_ in the last PR. It was the behavior that the `merge` operation had. 
(But I'll have to re-read whether this might be related to https://github.com/CesiumGS/3d-tiles-tools/pull/140/files#diff-41e63ef4640d275815adbe220f32436f140f03ff544b0804dc4bc58426d1fa46R153 ...?)

In any case, there certainly are different possible strategies for determining the URIs of the tilesets. And trying to set the URIs in the _target_ based on the URIs of the input and the output directory (using the `relativize` function) _could_ make sense in many cases. But it looks dangerous - insofar that there may be cases where this is not possible, does not make sense, or yields completely unexpected results. (I'm thinking about cases where someone uses `C:/some/absolute/path/to/some/tileset.json` as the input, or inputs and outputs that are in certain directory structures (relative to each other) where the `relativize` call bails out. 

I'll have to carefully read the changes, think this through more thoroughly, and maybe do some further tests. For example, with absolute paths (even though it could be tricky to put that into _unit_ tests...).

jo-chemla:
Thanks for getting back again. 
The original PR was not working correctly for input tilesets described by paths pointing to the json files `tilesetFolder/tilesetA.json`. Children URI was appended tileset.json twice and resulted in paths that looked like `tilesetFolder/tilesetA.json/tilesetA.json`

This new implementation works as it should for all cases I could test it on - see table below. The difference between how merge and mergeJson should handle tilesets filepaths is that the simple `merge` seems to find the best suitable output folderpath within the output merged tileset directory, while `mergeJson` should find the path relative to the output mergedTileset.json that points towards the input tilesetN.json. 

This new implementation was tested on multiple iterations of absolute and relative inputs and outputs. Children uris always result in the correct relative filepath - only case when it is an absolute path is when input and output are on different drive letters. Could not update the spec to handle test-cases with absolute paths, hence I only added the test for relative input and relative outputs filepaths. 

| inputs | output| Works? | Children uri|
|--------|--------|--------|--------|
| absolute `C:\tilesets\folderN\tilesetN.json` | absolute `C:\tilesets\merged\tilesetMerged.json` | ✅ | `../folderN/tilesetN.json` | 
| relative `.\tilesets\folderN\tilesetN.json` | relative `.\tilesets\merged\tilesetMerged.json` | ✅ | `../folderN/tilesetN.json` | 
| relative `.\tilesets\folderN\tilesetN.json` |  absolute `C:\tilesets\merged\tilesetMerged.json` | ✅ | `../folderN/tilesetN.json` | 
| absolute `C:\tilesets\folderN\tilesetN.json` | relative `.\tilesets\merged\tilesetMerged.json` | ✅ | `../folderN/tilesetN.json` | 
| absolute `C:\tilesets\folderN\tilesetN.json` | absolute on other drive `Z:\out\merged\tilesetMerged.json` | ✅ | `C:/tilesets/folderN/tilesetN.json` |

javagl:
(Only a short comment - I still have to allocate more time for that:) 

I had seen the different handling of the "identifiers" (already linked to above, [here in the last PR](https://github.com/CesiumGS/3d-tiles-tools/pull/140/files#diff-41e63ef4640d275815adbe220f32436f140f03ff544b0804dc4bc58426d1fa46R153)) that depended on the `jsonOnly` flag, and ... in hindsight, I should have looked at that with more scrutiny, to better understand why these identifiers should be handled differently based on the `jsonOnly` flag.

Some context: These "identifiers" are just used for disambiguating multiple input tilesets that "have the same name". This was not handled in the initial implementation of the `merge` function (waaay back when [this was still part of another repo](https://github.com/CesiumGS/3d-tiles-validator/commit/942ac8061b8e73574e0b3ee50cbf52406ee6d27c#diff-d3b04225b4056eb0078ddd75f7ed538a0e10f4dd6ae5d5d73b23c2f0ebc09f25R20)). And this case is what is explicitly checked with the current spec data, with the duplicate name appearing as `TilesetA` and `sub/TilesetA`. 

The problem:

The behavior of `merge` is still _horribly_ underspecified. It was just taken from the old state (linked above), revived, and fixed for the case of duplicate path components. But right now, the description - neither the CLI docs nor the TSDoc - say anything about the technical details, like what the resulting `content.uri` will be.

For example, the current (single) `merge` spec uses
```
inputs = [
  "./specs/data/mergeTilesets/basicMerge/TilesetA"
  "./specs/data/mergeTilesets/basicMerge/sub/TilesetA"
]
```
as the inputs, and tests that the resulting URIs for the tilesets will be
```
  "TilesetA/tileset.json",
  "TilesetA-0/tileset.json",
```
These `TilesetA` and `TilesetA-0` prefixes are totally unspecified. They could also be `randomStringX` and `randomStringY`, and **usually, this would not matter**, because these tilesets are just _copied_ to the output directory, into subdirectories with the proper names.

Now... with `jsonOnly`, they are **not** copied. And this affects the expectations that people might have for the resulting `content.uri`. The `merge` function can no longer assign random names (and disambiguate them with some unspecified `-0` suffix.

In the current state of the spec, you said
``` 
    const expectedContentUris = [
      "../../mergeTilesets/basicMerge/TilesetA/tileset.json",
      "../../mergeTilesets/basicMerge/sub/TilesetA/tileset.json",
    ];
```
Now, these content URIs are very long, very specific, and _very_ dependent on the exact structure of the input and output. And it **will** make a crucial difference of whether they use a directory structure like
```
input/TilesetA/tileset.json
input/TilesetB/tileset.json
->
output/tileset.json
```
or
```
inputTilesetA/tileset.json
inputTilesetB/tileset.json
->
output/tileset.json
```

From quickly looking over the table, it looks like the resulting `content.uri` do make sense. But I'll have to think this through (also in terms of disambiguation), and in how far the behavior here can be "specified" - maybe with some statement like

> The merged tileset will refer to the input tilesets using relative URIs that reflect the structure of the input and output directories..

(you know, roughly like that...). The fact that this is so important (only) for the `jsonOnly=true` case evaded me in the first PR.


---

Sorry... so much about a "short comment" 😬 But I see these PRs as a chance to clarify some aspects of the behavior and describe it more strictly on a technical level. And I appreciate your feedback, contributions, and tests that appear to happen based on real-world examples, and not only one artificial example that is run as part of a unit test.

jo-chemla:
Thanks for the "short comment", very much appreciated! It is now clearer to me, both the history behind the implementation and the use for these "identifiers" within the standard merge utility. 

I did understand indeed that these were made to disambiguate input tilesets directories named the same at two different locations, so they could be copied into output at the best location within output directory - as top-level as possible, with a suffix in case it is not the first children tileset directory named like that. 

In case of `jsonOnly=true`, the probable use-case is for the user to reference the output merged json tileset into a scene defined with CesiumJS or CesiumForUnreal etc. Having the children-uris of the referenced merged tilesets referenced by a path relative to the output merged tileset means that the tileset loaded into the scene will immediately correctly reference all children tilesets, and load the merged children where they are. Hence the new implementation - which does not modify identifiers for `jsonOnly=false`, but uses relative paths for `jsonOnly=true`. 

Hope this makes sense. In our workflow, this is the use-case we have, hence also testing it for all the different flavors of relative vs absolute input & output paths arguments to the CLI in the table above, to make sure all these cases are yielding expected results.

javagl:
When running the new spec, then it expects the content URIs to be
```
    const expectedContentUris = [
      "../../mergeTilesets/basicMerge/TilesetA/tileset.json",
      "../../mergeTilesets/basicMerge/sub/TilesetA/tileset.json",
    ];
```
However, the directory structure within `specs` is
```
// Inputs:
specs/data/mergeTilesets/basicMerge/TilesetA/tileset.json
specs/data/mergeTilesets/basicMerge/sub/TilesetA/tileset.json

// Outputs
specs/data/output/mergeTilesets/basicMerge/tileset.json
```
When I'm counting correctly, then there is one `../` missing in the URIs:

- Starting at `specs/data/output/mergeTilesets/basicMerge/` 
- First `../` goes to `specs/data/output/mergeTilesets/`
- Second `../` goes to `specs/data/output/`
- There is no `mergeTilesets` directory - it would need ...
- ... a third `../` to go to `specs/data/`
- There, it would find the `mergeTilesets` directory that contains the sources

Also, I kept the output files and served them with a local server, and the message
`A 3D tile failed to load: http://localhost:8003/output/mergeTilesets/basicMerge/sub/TilesetA/tileset.json`
seems to confirm this.

The reason for that is probably that the `path.relative` function behaves differently for _files_ and for _directories_ (i.e. it does not "know" whether something is a file or directory). When specifying the inputs _including_ the trailing `/tileset.json`, then it *does* insert the required third `../`. (I'll try to fix that locally...)

jo-chemla:
Thanks for the detailed feedback, very much appreciated. 
 - Completely agree that the second code-block looks way clearer, better differentiating between when `jsonOnly` is either true or false. 
 - I now realize that `tilesetTargetDirectoryPath` could have been taken out of the scope of the for loop. 
 - And indeed the third `..` is missing from the spec. The table above, on which I tested the new `mergeJson` implementation, was indeed only looking at inputs and outputs specified as filepaths, including `foldername\tilesetName.json`. I did not test it on directories since none of our tilesets are simply named `tileset.json` - I should have specified it, sorry for that! You are probably on the right track with that node `path.relative` not differentiating between filepaths and directories. 

TLDR: completely agree with the proposed changes, these look indeed better!

jo-chemla:
Thanks for the feedback, great to hear. @alouis-jpg from our team did test against the same 5 cases of the table with a mix of absolute and relative paths for inputs and outputs (only on tileset files since we have no tileset named tileset.json) and they all yielded the same childrzn uris with your PR. So this should be good to go!

javagl:
Based on your comments, this seems to now be resolved via https://github.com/CesiumGS/3d-tiles-tools/pull/143

Issue #165, opened 2024-12-27
The `upgrade` does not properly handle `gltfUpAxis`


javagl:
One legacy feature of 3D Tiles (which, I think, even predates 3D Tiles 1.0), was the possibility to specify an "up-axis" for the glTF assets that it referred to (indirectly, e.g. via B3DM). This was originally caused by the lack of an "up-axis" convention in glTF itself. 

Now, there may still be legacy tilesets that specify the `gltfUpAxis` like this:
```
  "asset": {
    "gltfUpAxis": "Z",
    "version": "1.0"
  },
```

Right now, the `upgrade` command does not handle this property at all. 

The `upgrade` command should check for this property, and 

- remove it from the `asset`
- take it into account when converting the B3DM to GLB

The latter is a bit tricky. It is technically simple, or even trivial: Just throw in a few axis conversion matrices here and there. But it is highly error-prone, due to certain corner cases. For example: It might be that a glTF 1.0 asset uses the `CESIUM_RTC` extension, or the batch table defines an `RTC_CENTER` (or... maybe even both 🤪 ). These RTC centers are already taken into account during the `upgrade`, and they already anticipate the Y-up-to-Z-up conversion of glTF vs. 3D Tiles. So one has to be very, _very_ careful to not mix up some axes when the `gltfUpAxis` has to be taken into account _in addition_ to that. The exact order of transformations will have to be derived from the (legacy) specifications (or maybe even from the CesiumJS implementation), and verified with dedicated spec data sets.

A **very** quick draft of handling the case that the input contains `gltfUpAxis=Z` is in the branch [`up-axis-handling-draft`](https://github.com/CesiumGS/3d-tiles-tools/tree/up-axis-handling-draft), with all relevant changes in the commit https://github.com/CesiumGS/3d-tiles-tools/commit/b5dab256784c78ac13879fe030a4cb32b78280f0 . But this makes the _assumption_(!) that the input has `gltfUpAxis=Z`, and is **not** a general solution.

javagl:
This was resolved via https://github.com/CesiumGS/3d-tiles-tools/pull/166 . There may still be obscure combinations of legacy features where ~"something does not work as desired", but these should be tracked in own issues if they come up.

PR #166, opened 2025-01-08
Take glTF up axis into account during upgrade

javagl:
Addresses https://github.com/CesiumGS/3d-tiles-tools/issues/165 

What this currently does:

- It checks for the presence of the `tileset.asset.gltfUpAxis`
- If it is found, then it is removed from the tileset JSON, and passed on the the tile content upgrades
- Until now, this is *only* used for the conversion of B3DM with glTF 1.0 into glTF 2.0 (when upgrading to 1.1)

It could be desirable to also use this for the `upgrade` to 1.0: The `gltfUpAxis` is a "legacy" feature that should not be part of a tileset that has been upgraded to 1.0. And this might be doable with reasonable effort. 

What I'm more scared of is that the `gltfUpAxis` also (likely) will affect the I3DM upgrade. The coordinate system conversions for I3DM had been a hassle. The coordinate system conversions for the `gltfUpAxis` are a hassle. Combining this will be hassle², with the caveat that this involves generating that "legacy" data in the first place...

The current test data (which is not yet checked in, and might still change) is here:

[upAxisHandling 2025-01-08.zip](https://github.com/user-attachments/files/18353449/upAxisHandling.2025-01-08.zip)

It focusses on the B3DM case (with and without an `RTC_CENTER` in the B3DM), and makes sure that the visual appearance of the output is still the same as the input in CesiumJS (Sandcastle and output are included)

Cesium Axes 2025-01-08

This is supposed to eventually become a test data set for the specs, using the "golden" comparison approach.


javagl:
The last commits consisted of cleanups, adding the same functionality for I3DM, and creating/adding the corresponding test data as specs.

The specs use the "golden" approach: The [`upgradeTileset/upAxisHandling`](https://github.com/CesiumGS/3d-tiles-tools/tree/cabdedb51205b0237514345d5b5563f052cee38a/specs/data/upgradeTileset/upAxisHandling) directory contains the `input` directory, with the various test cases, and the `golden` directory with the _expected_ output. When the `TilesetUpgraderAxesSpec` is run, then it will write the results into an `output` directory and compare them to the `golden` one.

The directory also contains a short README and a Sandcastle for comparing the input and output. 

The case of `gltfUpAxis`+I3DM indeed brought up some issue, but ... one can argue whether this is something that should be handled: When the glTF 1.0 in an I3DM uses Z- or X-up, and this is **not** declared via the `gltfUpAxis` in the tileset JSON, then the upgrade does not generate a proper (unmodified) GLB. I'd have to spend much more time (with creating test data, and probably pen+paper) to figure out _whether_ this can sensibly be addressed, and how: The conversion of the I3DM into `EXT_mesh_gpu_instancing` inevietably _has_ to make assumptions about the meaning of the translations in the I3DM, and how they should be transferred into the glTF extension. 

The following shows the Sandcastle for comparing the inputs and outputs, pausing for a moment in this case:

Cesium Up Axis Handling Sandcastle 0001

I think that this can be justified: The upgrade cannot magically "know" how the data was _supposed_ to be interpreted. And when the `gltfUpAxis` **is** specified, then the upgrade works as expected. 

---

Beyond that, there may still be a million cases that don't work. Maybe some tileset with `gltfUpAxis=Z` that refers to _external_ tilesets with `gltfUpAxis=X`? Or obscure combinations of the `CESIUM_RTC` extension and an `RTC_CENTER` in the B3DM? ...

What I _did_ test, though, was the current state with a user-provided real-world data set that used `gltfUpAxis=Z`, **and** the `CESIUM_RTC` extension, and I made sure that the result is the expected one. So even when there may be many cases where ~"something that may appear in legacy data" does _not_ work, at least one real-world configuration of legacy data _can_ be properly upgraded now.

Issue #159, opened 2024-11-24
Consider other solutions for `Iterables`...


javagl:
The current [`Iterables`](https://github.com/CesiumGS/3d-tiles-tools/blob/0e8f5e082f7811b7fd81bc34a3d1d1646125522c/src/base/base/Iterables.ts) class has been created to offer basic utility functions for iteration. These should have been part of the core language specification since the `Iterator` concept was established, but ... haven't. Given some possible upcoming refactorings (e.g. in the context of https://github.com/CesiumGS/3d-tiles-tools/issues/156 ), the same functionalities will also be required for _asynchronous_ iterators. 

There are several options for the "other solutions" mentioned in the title:

- Replacing that class with a library dependency
- _Extracting_ it _into_ a library/package
- Replacing it with built-in functionalities

The last one would be the most desirable. And it looks like in the meantime (roughly a month ago), https://github.com/tc39/proposal-iterator-helpers has been brought on its way to become part of the core standard (but who knows how long _that_ will take). That won't be enough, though:  We'd still have to wait for https://github.com/tc39/proposal-async-iterator-helpers ...

There certainly are dozens of existing libraries that offer similar functionality, and much (**MUCH**) more that could be useful here. (and ... are better tested). Some of them are listed in https://github.com/tc39/proposal-iterator-helpers?tab=readme-ov-file#prior-art--userland-implementations . But a built-in solution would be preferable in many ways (particularly because it's not clear when and how these libraries will catch up with asnychronous iteration protocols...)

---

Note: When looking at the functions in `Iterables`, one might wonder why they are written in this seemingly complicated way of always creating a `resultIterable ... [Symbol.iterator] ..`. Some of the "boilerplate" code there could be avoided (with some `private static createIterator(...)` function or so). But the underlying issue here is that a JavaScript `Generator` concept is ... well, I'll just call it "broken": It is an itera**tor** that looks like an itera**ble**. But it is itera**ble** only **once**:

```javascript
function* create() {
  yield "ELEMENT0001";
  yield "ELEMENT0002";
  yield "ELEMENT0003";
  yield "ELEMENT0004";
}

function example() {

  const iterable = create();

  console.log("Let's iterate, because it's iterable");
  for (const element of iterable) {
    console.log(element);
  }

  console.log("Let's try that again, because it's iterable");
  for (const element of iterable) {
    console.log(element);
  }
  console.log("Whoopsie, why didn't it show anything?");
}

example();
```

[...]

Issue #156, opened 2024-11-18
Support for asynchronous tileset sources (and targets)


javagl:
The tools can currently process tilesets in different representations. The most common one is that of a tileset being stored in the file system, with some `tileset.json` that essentially points to the content files using relative paths in the `content.uri`. But they also support tilesets that are stored in 3DTILES files (an SQLite-based database file) or in 3TZ (a ZIP-based storage format).

For the internal processing of the tileset data, these different representations are completely abstracted away: All the operations take place on a [`TilesetSource`](https://github.com/CesiumGS/3d-tiles-tools/blob/d549136b9edb5af0a8a6ea105e2e87d782204d07/src/tilesets/tilesetData/TilesetSource.ts) and a [`TilesetTarget`](https://github.com/CesiumGS/3d-tiles-tools/blob/d549136b9edb5af0a8a6ea105e2e87d782204d07/src/tilesets/tilesetData/TilesetTarget.ts). These are interfaces that offer the basic operations that are required for implementing the tools functionality.

Right now, nearly all methods in these interfaces are _synchronous_. This makes sense for the existing implementations. But it limits the extensibility: If there ever should be a representation of a tileset where any aspect is _asynchronous_, then this cannot be accomplished with these interfaces. 

It could therefore make sense to change this, and to declare **all** functions of the `TilesetSource` and `TilesetTarget` as `async`. The changes on the usage side should be minor, insofar that it can be handled by just slamming an `await` before each call. 

---

A rant:

I personally don't like that 'async' is _viral_. When there is a call chain like
```
function getA() {  return computeA(getB()); }
function getB() {  return computeB(getC()); }
...
function getZ() {  return someLibrary.getZ(); }
```
and that `someLibrary` is later swiched to a library that only offers `getZ` as an `async` function, then the whole thing breaks down, and one has to pass that `async/await` though the whole call chain, in a completely breaking change.

This makes **so** little sense that I thought that there should be a programming language that could just be called 'AsyncScript`, where each and every function call like
```
function example() {...}
const result = example();
```
is implicitly and automatically translated into the JavaScript code
```
async function example() {...}
const result = await example();
```
(And making this translation 'smart and efficient(!)' should be the job of the compiler)

But given the absence of such a language, my stance is: Functions in basic interfaces should be declared as `async` _by default_, because that is _literally_ the only way to be on the safe side and avoid breaking changes with future implementations.

javagl:
In this context: We should consider adding
[...]

PR #167, opened 2025-01-19
Make tileset sources and targets asynchronous

javagl:
Addresses https://github.com/CesiumGS/3d-tiles-tools/issues/156 

The main part of the changes here is purely syntactical: Just add that `async` to each function in the [`TilesetSource`](https://github.com/CesiumGS/3d-tiles-tools/blob/4889d7f425f1b7b7bfe8baaf6fb97d611a9ff851/src/tilesets/tilesetData/TilesetSource.ts) and [`TilesetTarget`](https://github.com/CesiumGS/3d-tiles-tools/blob/4889d7f425f1b7b7bfe8baaf6fb97d611a9ff851/src/tilesets/tilesetData/TilesetTarget.ts), and the corresponding `await` ... _everywhere_ (which explains the many "Changed files"). 

There are also some smaller cleanups and utility functions. Some "slightly larger" changes are in the [`Iterables`](https://github.com/CesiumGS/3d-tiles-tools/blob/4889d7f425f1b7b7bfe8baaf6fb97d611a9ff851/src/base/base/Iterables.ts) class: This provides various utility functions for `Iterable` objects. Many of them had to be extended/generalized to `AsyncIterable`, because things like the `tilesetSource.getKeys()` function now returns an `AsyncIterable`. Some further refactoring may happen there. Maybe all this can be replaced by some nice library. Further thoughts are in https://github.com/CesiumGS/3d-tiles-tools/issues/159 

Another addition was to bring "iteration" and "traversal" closer together: The [`TilesetTraversers`](https://github.com/CesiumGS/3d-tiles-tools/blob/4889d7f425f1b7b7bfe8baaf6fb97d611a9ff851/src/tilesets/traversal/TilesetTraversers.ts) class now offers a few functions to create `AsyncIterable` objects over the traversed tiles. (There are a few degrees of freedom and ways to extend or generalize this, but the current state is a baseline).

One thing that is ... _somewhat_ unrelated to the core of the PR, and where I'm not sure whether it should be included here or added in a dedicated PR: Many of the functions that operate on tilesets receive the `tilesetSourceName` and `tilesetTargetName` (e.g. `"C:/example/tileset.json"`). This is intentional, and offers convenience for most use-cases. But internally, _most_ of these functions are doing the same thing: They are creating a `TilesetSource`/`TilesetTarget` from the given names, and then do their job on these objects. In order to allow clients to use custom implementations of these interface, there will have to be functions that do not accept strings (i.e. not the names), but the `TilesetSource`/`TilesetTarget` objects _directly_.


javagl:
>  Many of the functions that operate on tilesets receive the `tilesetSourceName` and `tilesetTargetName` [...] In order to allow clients to use custom implementations of these interface, there will have to be functions that do not accept strings (i.e. not the names), but the `TilesetSource`/`TilesetTarget` objects directly.

This is now part of this PR (mainly done in https://github.com/CesiumGS/3d-tiles-tools/commit/e1afcb4be01b0f3bd95c00836371da9ceb5fde2c ). This means that for the existing functions (like `combine`, `merge`, `upgrade`...) that received the source- and target names (as strings), there are now corresponding functions (`combineData`, `mergeData`, `upgradeData`...) that take the source- and target data (as `TilesetSource` and `TilesetTarget` objects) directly.

There are no additional implementations of these interfaces yet. But future implementations now _can_ be asynchronous, and that was the main point here.

Issue #170, opened 2025-02-06
Upgrade - Error: Missing required extension, "CESIUM_RTC".


sweco-sekrsv:
I'm trying to upgrade tileset exported from FME Form. 

This is the command I use:
`npx 3d-tiles-tools upgrade -i "C:\Users\sekrsv\Downloads\6_PROJEKTERING SH\upgrade\tileset.json" -o "C:\Users\sekrsv\Downloads\6_PROJEKTERING SH\upgrade2" --targetVersion 1.1`

I get this error repeated for all the files:
Failed to upgrade data/59/data159.b3dm: Error: Missing required extension, "CESIUM_RTC".

I have attached an example b3dm file.

Is there anything I can do to fix this or is this something that you need to look at and add a fix for to the library?

[data159.zip](https://github.com/user-attachments/files/18687676/data159.zip)

javagl:
Just to quickly acknowledge that: It does indeed cause an error. The reason is that [`CESIUM_RTC`](https://github.com/KhronosGroup/glTF/tree/main/extensions/1.0/Vendor/CESIUM_RTC) is an extension for glTF **1.0**. But here, it appears in a glTF **2.0** that is contained in this B3DM. On top of that, the extension is marked as 'required' by appearing in `extensionsRequired`. (It should not really be required, but the specification does not point that out explicitly). All this together causes glTF-Transform to throw up when trying to read that glTF, because it claims that the extension is required, and glTF-Transform does not support this (glTF 1.0) extension.

I've occasionally seen people casually putting `CESIUM_RTC` into glTF 2.0 (and pointed out the caveats with that). But I wasn't aware that this is apparently used by some widely used tool. 

It would be nice if the 3D Tiles Tools could handle this better. The functionality for converting the extension into a node transform already exists for the upgrade from glTF 1.0 to glTF 2.0. It's a bit unfortunate that it will be necessary to "inspect and fix" _every_ glTF just for the case that it _might_ contain this extension. But I'll have a look at how this could be addressed.

javagl:
I created a PR that is supposed to fix this, at https://github.com/CesiumGS/3d-tiles-tools/pull/172 

I added a test file there (i.e. a B3DM with glTF 2.0 that contains the extension), and it seems to work as expected. Maybe you want to try out this state on your tileset as well, just as another test case.

PR #172, opened 2025-02-06
Handle CESIUM_RTC in glTF 2.0

javagl:
Fixes https://github.com/CesiumGS/3d-tiles-tools/issues/170 :  

The [`CESIUM_RTC`](https://github.com/KhronosGroup/glTF/tree/main/extensions/1.0/Vendor/CESIUM_RTC) extension is an extension for glTF **1.0**. But there are some glTF **2.0** assets in the wild that contain this extension. This is somewhat problematic: Nearly no tool or viewer will support this extension. 

However, when a glTF 2.0 asset with `CESIUM_RTC` is encountered in a B3DM during the `upgrade`, then the 3D Tiles Tools should definitely be able to handle this. This is accomplished with this PR.

There already _was_ a functionality to change the `CESIUM_RTC` into a translation of the root node when upgrading from glTF 1.0 to 2.0. with `gltf-pipeline`. So the change here is trivial: This step is also done when the input already _was_ glTF 2.0.

---

**EDIT:** The following largely became obsolete. There are several relevant commits now, and the changes in the specs should reflect the extended and updated functionality as described in the comments below.

---

The only change that really has to be reviewed is https://github.com/CesiumGS/3d-tiles-tools/commit/9b23d038126b5ba32e496dfdd91d554fc3675f7c 

Why are there 54 modified files then? 🤕 

First of all, some parts of the "upgrade" functionality had not been covered with specs. There have been some example tilesets that have been used in the "demos", but not ran as part of the specs. These example tilesets are now run as part of the specs, and I added another one, `tilesetWithB3dmWithGltf2WithCesiumRtc`.

The test approach is the one that is commonly used in other places, namely that of the "golden" output. As usual, there is a sandcastle included, for comparing the input and output:

Cesium TilesetUpgraderSandcastle

This _still_ does not explain _all_ of the changed files 😬 

I also had to update some of the other "golden" outputs. They have changed _slightly_. After a bit of headscratching, I realized that the differences come from the fact that `gltf-pipeline` is now touching glTF 2.0 files that it previously did _not_ touch, and ... for some reason, `gltf-pipeline` [assigns the name `"buffer"` to the main buffer](https://github.com/CesiumGS/gltf-pipeline/blob/cbbff2f212b5aa6163c7152d46ffcaf3a2ff3ac3/lib/mergeBuffers.js#L29). So... many of the changes here are just reflecting that.


javagl:
This will be updated to 
- replace the `CESIUM_RTC` extension _manually_ (not with `gltf-pipeline`), to reduce the likelihood of accidental structural changes
- be able to apply the `upgrade` also to GLB/GLTF file content (right now, it was only applied to B3DM/I3DM - but here, the necessity to remove that extension also applies to content that already **is** glTF)

javagl:
The last commit changes the handling of the `CESIUM_RTC` extension as discussed internally.

- Previously, it did pass the GLB data through `gltf-pipeline`, and **iff** the data contained `CESIUM_RTC`, then it was replaced. This could lead to unexpected (or unnecessary) changes (like the `buffer.name` being inserted)
- Now, it explicitly checks the JSON from the GLB, looks whether it _does_ contain the `CESIUM_RTC`, and _only_ replaces it (without touching or modifying any other part of the JSON)

(This changed the "golden" spec files accordingly. Previously, there had been 54 modified files, due to the insertion of the `buffer.name` by `gltf-pipeline`. Now, these changes have basically been rolled back)

javagl:
Back at 53 files 🙂 these now include specs for the latest change here: The upgrade will now also examine `.gltf` and `.glb` tile content (which is always glTF 2.0). If this tile content still contains the `CESIUM_RTC` extension, then it will be upgraded accordingly. 

As part of this, I made the `gltfUpAxis` a _required_ parameter throughout the functions. This parameter was introduced as part of https://github.com/CesiumGS/3d-tiles-tools/pull/166 , but was marked as optional, as in `gltfUpAxis?: ...`. I usually try to avoid optional parameters, _exactly_ because of what happened here: It did lead to cases where it was accidentally omitted. It's ... not s nice ... that this now has to be passed along through all the functions, but it always affects the handling of the `CESIUM_RTC` translation, so it is crucial that it is _really_ passed in everywhere.

PR #173, opened 2025-02-08
Fix handling of local file header for 3TZ

javagl:
There have been reports of 3TZ files that could not be read due to improper handling of the "extras" field in local file headers. The 3TZ reading functions simply assumed that the "extras" are at most 48 bytes (and I don't know where this value came from). But certain files that had been accessed with certain tools could contain extras that contained _more_ than 48 bytes (in a ["Windows NT Security Descriptor Extra Field", see specs](https://libzip.org/specifications/appnote_iz.txt)). 

The change here removes the assumption of these 48 bytes. The archive functions now explicitly look at the _actual_ size of the extras field in the header, and then use this to compute the offset at which the _actual_ file data can be read. 

I'm opening this as a "draft" for now. I'd like to add some tests/specs. But in terms of 3TZ, there currently is _zero_ test coverage, and I don't know how the problematic files had been created, so it's not entirely clear how the test coverage could sensibly be increased here...


javagl:
While creating specs for this, I noticed that there is another hard-wired (wrong) constant [of 320 bytes](https://github.com/CesiumGS/3d-tiles-tools/blob/a07ffdf1ad93591954a9a98936f7cf7b6617f4d9/src/tilesets/packages/ArchiveFunctions3tz.ts#L40) that was carried over from the [original implementation](https://github.com/bjornblissing/3d-tiles-validator/blob/2f4844d5bdd704509bff65199898981228594aaa/validator/lib/archive.js#L34), meaning that the functions cannot read 3TZ files that are smaller than 320 bytes (like the one that I just created for the specs 😬 ). I'll add a fix for this as well...

javagl:
... and this fix was pretty trivial: Just don't try to read more than the file size, as in 
``` 
- const bytesToRead = 320;
+ const bytesToRead = Math.min(320, stat.size);  
```

I also added some spec files for the 3TZ reading functionality that _explicitly_ covers these corner cases. (There are other specs that causally deal with 3TZ, but that do not cover some of these low-level details)

javagl:
I _think_ that this could be ready for review, but I'll use this state more extensively for tests in the `3d-tiles-validator` (c.f. https://github.com/CesiumGS/3d-tiles-validator/pull/331#issuecomment-2654490400 ) and if there are any further 3TZ related issues in the tools, I'll fix them as part of this PR.

javagl:
I tested this more extensive, in the context of https://github.com/CesiumGS/3d-tiles-validator/pull/332 , with real-world tilesets that contain (many, huge) 3TZ files, and everything seemed to work properly. So this should be good to go.

PR #180, opened 2025-05-27
Added the rotation option to createTilesJson tool

SagarB-97:
Currently, the createTileJson option only takes cartographic position in degrees as input. I have added the option to also accept heading, pitch, and roll in degrees.

### DIFF
--- a/src/cli/ToolsMain.ts
+++ b/src/cli/ToolsMain.ts
@@ -598,6 +598,7 @@ export class ToolsMain {
     inputName: string,
     output: string,
     cartographicPositionDegrees: number[] | undefined,
+    rotationDegrees: number[] | undefined,
     force: boolean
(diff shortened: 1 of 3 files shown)


javagl:
Thanks for that contribution. As a feature, I think that this makes sense and increases the flexibility of the `createTilesetJson` command. I'll try to allocate some time for a closer review and test. 

In the changed files (and by the failed build run), I noticed some formatting changes that are unrelated to the main change. When you run
`npm run prettier`
then it should format the code according to what is predefined via the ESLint config (and then, the build should likely pass).

SagarB-97:
Thanks for the response. The formatting changes were made automatically, and I didn’t notice them since the diffs weren’t visible in my IDE. I’ll update it.

Issue #127, opened 2024-05-08
Update to glTF-Transform 4.0


javagl:
glTF-Transform 4.0.0 has recently been published.

We should consider an update soon. The first pass would be to "just update the version number and see whether the compiler complains" (and then see whether the tests are still passing). 

I haven't studied the change log yet. Unless there are severe, _conceptual_ breaking changes, I wouldn't expect too much hassle here - _except_ for the implementations of the metadata extensions. These are "closer to the guts" (internal implementation details) of glTF-Transform than the other parts of the tools, and might therefore be affected even by more subtle changes.

---

One thing that has to be updated and can therefore be tracked here: In the context of upgrading CMPT to GLB, the corresponding PR derailed into one that started to cover the (huge and open) topic of "merging metadata". As part of that, some functions from glTF-Transform 4.0 have essentially been _copied_ to [an internal utility class](https://github.com/CesiumGS/3d-tiles-tools/pull/117/files#diff-16e9253b0431394c8cf714e4fb0dca213bd4c3e93e6a7e0a5a61d7d6ce6c56eeR1) (with notes about the source for attribution, of course!). These should be replaced by the built-in functions after the update.

LoongZP:
Hello, as a developer who is using it, I am eager to know when it will be upgraded to glTF-Transform 4.0.0. Now I'm trying to upgrade it.

javagl:
I have not yet investigated the details here. For the main part, I assume that the upgrade _should_ be simple, assuming that there are not sooo many breaking changes in 4.0.0 that are actually relevant for the tools. 

But there is one place that requires special attention here: The [`StructuralMetadataMergeUtilities`](https://github.com/CesiumGS/3d-tiles-tools/blob/b8bad9ee568d5570b7c711411147d932e6b0c5df/src/tools/gltfExtensionsUtils/StructuralMetadataMergeUtilities.ts) have been "ported" from glTF-Transform 4.0 to work around some limitations of 3.x, and this will have to be tested and adjusted (and preferably simply be _removed_) eventually.

I'm curious, though: In how far is the actual glTF-Transform version that is used in the tools relevant for you? (At the command line, it should not matter at all, and on the library level - which is still not officially stable, with the tools being in a 0.x version - there should not be many differences either...)

javagl:
[...]

Issue #181, opened 2025-05-30
The reading order of EXT_structural_metadata, EXT_mesh_features, and EXT_instance_features.


LoongZP:
I used convertB3dmToGlb to convert the b3dm file to glb, but I cannot read it with gltf-transform. The error report is as follows:

![Image](https://github.com/user-attachments/assets/8eefe580-ea75-430b-9162-acbd92789817)

The reason is that the read function of EXT_mesh_features executes before that of EXT_structural_metadata, but the read function of EXT_mesh_features depends on the results read by EXT_structural_metadata. As shown in the image below:

![Image](https://github.com/user-attachments/assets/a2eff91a-a3c6-445e-85ed-e070fa8cd79d)

https://github.com/donmccurdy/glTF-Transform/issues/1659#issuecomment-2919250018 It is recommended to use the preread function.

LoongZP:
Sorry, I haven't looked into the code of GLTFReader in gltf-transform in detail, and it seems that preread cannot solve this problem.

javagl:
I assume that this **only** happens with glTF-Transform 4.0, and you are encountering this in the context of https://github.com/CesiumGS/3d-tiles-tools/issues/127 . 

And coincidentally, this issue is caused by a change that was introduced because of [an issue that I brought up](https://github.com/donmccurdy/glTF-Transform/issues/1256). And in that issue, @donmccurdy already said

> This will have to go into v4 rather than v3.x, as the sorting (at least) is potentially a breaking change.

And we now have such a case.

I'll have to take a closer look at how this can be resolved. If you gather any insights or options, then they can be tracked here.


---

An anside:  I'm usually using the [`tools/contentProcessing/GltfTransform.getIO()` function](https://github.com/CesiumGS/3d-tiles-tools/blob/bcf307301c776a47f1ec7f055bd28b0a1ac10415/src/tools/contentProcessing/GltfTransform.ts#L56) to obtain a glTF-Transform IO object that is conveniently preconfigured with all decoders and extensions, and this function contains a note about that issue:
```javascript
    // Note: The order of these calls matters. The EXTMeshFeatures and
    // EXTInstanceFeatures depend on the EXTStructuralMetadata, because
    // they may refer to PropertyTable objects via their FeatureId.
    // So the EXTStructuralMetadata has to be read first.
    io.registerExtensions([EXTStructuralMetadata]);
    io.registerExtensions([EXTMeshFeatures]);
[...]

PR #185, opened 2025-06-05
Update to glTF-Transform 4

javagl:
Addresses https://github.com/CesiumGS/3d-tiles-tools/issues/127 

Opened as a draft PR for now.  Details will be added here soon.


src/gltf-extensions/gltfExtensions/EXTStructuralMetadata.ts
donmccurdy: Addressed in https://github.com/donmccurdy/glTF-Transform/pull/1810 already but for FYI, I suggest moving this implementation from `read()` to `preread()` registered with `prereadTypes: [PropertyType.SCENE]`. That is the last hook available before read() will be called on all remaining extensions, allowing EXT_structural_metadata to be initialized before EXT_mesh_features.

donmccurdy:
Work in progress:

- https://github.com/donmccurdy/glTF-Transform/pull/1810

javagl:
Great! Most of the open questions that prevented this PR from being finalized/merged revolved exactly around the points that will become obsolete when these extensions are actually part of glTF-Transform itself.

The first point was about the behavior of `copyToDocument` and `mergeDocuments`. As mentioned above, this seemed to work smoothly, given that they had essentially been "copied" from glTF-Transform anyhow.

Other points already came up when writing the original implementation. And they may be unrelated to the update to 4.0.0 in particular. They are rather about "best practices", and where _exactly_ parts of the implementation should go. Specifically, these questions have been related to ...

1. `EXT_structural_metadata` being 'very top-level' and requiring an early initialization
2. The order of the extensions, and the dependency to `EXT_mesh_features`
  - Related to https://github.com/donmccurdy/glTF-Transform/issues/1256 , where you already gave some hints about `read/preread`
  - Related to https://github.com/CesiumGS/3d-tiles-tools/issues/181, which may be closed/obsolete soon
3.  The fact that `EXT_structural_metadata` uses _buffer views_, which don't really exist in glTF-Transform.

The last one was solved with `prewriteBuffers` (I think that was already based on some hints that you gave me back then). I did have to dive a bit into the implementation details of glTF-Transform, and was not sure whether this [`obtainOtherBufferViews` call in `prewriteBuffers`](https://github.com/CesiumGS/3d-tiles-tools/blob/d48152cb25fecb7c13b63745497de7c5151421b2/src/gltf-extensions/gltfExtensions/EXTStructuralMetadata.ts#L1374) and collecting these "other buffer views" in the context was really the most idiomatic way to solve this. (It seemed to work, and make sense, but I wasn't sure...)

Now, I'm not entirely sure about the timeline or order of operations. The update here is not time-critical, but has been pending for far too long. Depending on the intended timeline for the support of these extensions in core glTF-Transform, I could try to wrap up _this_ PR first, or _remove_ the extension implementations here as part of an update to glTF-Transform 4.x.0, with 'x' being whatever version this will be part of.

src/gltf-extensions/gltfExtensions/EXTStructuralMetadata.ts
donmccurdy: Ran into a typings issue here, to be solved by:

- https://github.com/donmccurdy/property-graph/pull/209

javagl: From a quick look, I don't see where this came up and what the 'issue' was. 

This type only appears in a `...Def`. And I assumed these to be ~"some JSON" which I expected to not affect the "internal type management" of glTF-Transform or property-graph. (But it looks like an easy fix in property-graph, and hopefully acceptable - assuming that further nestings, like `number[][][]`, `number[][][][]`, ... etc are unlikely to come up in practice...)

In the "model"-representation, i.e. in the `ClassProperty extends ExtensionProperty`, these types currently resort to `any` anyhow. I probably considered it back when I implemented this, but am now wondering if that could be improved, i.e. if these `any`s could be made more specific...

donmccurdy: Yeah, my intention was to replace `any` here ...

https://github.com/donmccurdy/glTF-Transform/pull/1810/changes/a5e29a3884caedf027c3d5702735de0823cd73fa

... at least in the glTF Transform codebase, these .get() / .set() methods require Literal / LiteralKey values for the `property-graph` type inference to work correctly.

donmccurdy:
Published in glTF Transform v4.4.0-alpha.1 —

- https://github.com/donmccurdy/glTF-Transform/pull/1810

Nominal changes made to match naming and typing conventions (e.g. `FeatureId` -> `FeatureID`) and to reflect later changes to the specification in https://github.com/CesiumGS/glTF/commit/d6958e16b103b6c5d588728351f56b5442121466.

javagl:
Great! I'll hopefully be able to actually _do_ wrap up this PR (which has been pending for nearly a year now) rather sooner than later. It looks like the extensions implementation package can be removed, and most of the updates in the `...Utils` package should already be done.

javagl:
@donmccurdy  You may have seen the note about the `name` property in an [inlined comment](https://github.com/CesiumGS/3d-tiles-tools/blob/d48152cb25fecb7c13b63745497de7c5151421b2/src/gltf-extensions/gltfExtensions/StructuralMetadata.ts#L24). I've seen that the "name" in these classes is now basically falling back to the (built-in) "name" property (which _usually_ is only in the [`glTF` **`ChildOfRoot`**`Property`](https://github.com/CesiumGS/glTF/blob/d757b9c77fe86b5a63f6514aa24f3de801effa7e/specification/2.0/schema/glTFChildOfRootProperty.schema.json)). Until now, this doesn't seem to cause issues. (I probably should have added more details back then, about what "conflicts" means - maybe calling it `objectName` was just an attempt to avoid ambiguities that _wouldn't_ have been an issue to begin with...)

---

The original implementation had this separation of `createSomething` (which are the way of creating sub-objects), and then split them into these [`createSomethingFrom`/`readSomething`](https://github.com/CesiumGS/3d-tiles-tools/blob/d48152cb25fecb7c13b63745497de7c5151421b2/src/gltf-extensions/gltfExtensions/EXTStructuralMetadata.ts#L323) pairs. I thought that this was a nice, clean way of separating the concerns here, and the `createSchemaFrom` was useful in the `StructuralMetadataMerger`. But of course, when this is supposed to be part of the glTF-Transform "core" functionality, then these functions should not be part of the public API surface. (I'll have to replicate some functions for the `StructuralMetadataMerger`, but nothing toooo big...)

---

It looks like [`Schema#set/getSchemaUri` assumes the schema URI to be non-`null`](https://github.com/donmccurdy/glTF-Transform/blob/3a8059a6ca21ca1cda23b730ef28b5e1deff63d2/packages/extensions/src/ext-structural-metadata/metadata.ts#L167), but it _can_ be `null`. Thanks to the anarchy of TypeScript, an `as any` resolves all issues here 🙂 Some places required changing
`if (schemaUri !== null)`
to
`if (schemaUri !== null && schemaUri.length !== 0)`
though...

---

I was not (and still am not) sure about the best way of handling "defaults", **consistently** in a library that provides thick abstraction layers for JSON like glTF. This refers to glTF-Transform as well as JglTF. I've seen the `getDefaults()` method in the `ExtensionProperty`, and _tried_ to apply it consistently. But there probably have been a few places in other parts of the 3D Tiles Tools where the `arrayOffsetType` and `stringOffsetType` properties had been set to their default, `UINT32`, as some sort of "convenience", so that users wouldn't have to deal with the `undefined` everywhere. 

Long story short (too late...) : 

- I had introduced this maybe-not-so smart `ifNot` function to omit some of the values in the `Def` structures when they had their default values in the "model", structures, e.g. [for the `arrayOffsetType`](https://github.com/CesiumGS/3d-tiles-tools/blob/d48152cb25fecb7c13b63745497de7c5151421b2/src/gltf-extensions/gltfExtensions/EXTStructuralMetadata.ts#L1225)
- In the current state in glTF-Transform, it seems to set them more consistently to _whatever is there_ - like the [`arrayOffsetType`](https://github.com/donmccurdy/glTF-Transform/blob/3a8059a6ca21ca1cda23b730ef28b5e1deff63d2/packages/extensions/src/ext-structural-metadata/structural-metadata.ts#L729) now, meaning that these defaults are part of the output, in the cases where the 3D Tiles Tools classes have been setting them _explicitly_ 

I now started to accept `undefined` for the `...OffsetType` properties more consistently, but that may require another cleanup pass. (There's also a bunch of differences in the "golden" spec files for the migration specs on top of that, but ... these are mainly due to the _order_ of the properties. I have to verify them, but that shouldn't be toooo much effort)

donmccurdy:
> ... maybe calling it `objectName` was just an attempt to avoid ambiguities that wouldn't have been an issue to begin with.

It looked to me like this should be safe, glTF Transform doesn't limit the "name" property to child-of-root properties if extensions declare it. Wouldn't be a problem to restore some special treatment of the property if needed, I just didn't want to do that unless/until there was a unit test identifying the unsafe scenario, and it looked safe as-is.

> The original implementation had this separation of createSomething (which are the way of creating sub-objects), and then split them into these [createSomethingFrom/readSomething](https://github.com/CesiumGS/3d-tiles-tools/blob/d48152cb25fecb7c13b63745497de7c5151421b2/src/gltf-extensions/gltfExtensions/EXTStructuralMetadata.ts#L323) pairs...

Yes, I stripped this down a bit for parallelism with the other extensions – hoping that's not too burdensome, but if more public interface is needed then I think we could work that out. `KHR_xmp_json_ld` is a bit similar here - the extension is a bit more complex but glTF Transform currently exposes only a minimal API surface to read/write.

> It looks like [Schema#set/getSchemaUri assumes the schema URI to be non-null](https://github.com/donmccurdy/glTF-Transform/blob/3a8059a6ca21ca1cda23b730ef28b5e1deff63d2/packages/extensions/src/ext-structural-metadata/metadata.ts#L167), but it can be null...

This one is intentional, and handled similarly for properties like "name"... to avoid ambiguity around `"" | null | undefined` glTF Transform mostly uses empty strings to represent any nullish/undefined/empty string input. When serializing to disk the empty strings are interpreted as undefined as needed. If some extension required a different meaning between empty strings and null or undefined I suppose that would need to change, but this has kept the types simpler elsewhere in the library so far.

> I was not (and still am not) sure about the best way of handling "defaults", ...

I've generally tried to avoid writing 'default' values into JSON output, to save on size, but that's not automated and probably isn't 100% consistent in glTF Transform, yeah. I'm comparing snapshots of sample files before/after a round-trip through glTF Transform in this test, that did catch a number of places where glTF Transform had been previously introducing explicit output of defaults:

https://github.com/donmccurdy/glTF-Transform/blob/main/packages/extensions/test/structural-metadata.test.ts

package.json
```diff
@@ -85,13 +85,13 @@
     "about:package-clean": "echo 'Delete the contents of the 'build' output folder'",
     "about:package-prepare": "echo 'Prepare a package: Linting, formatting, build, test, documentation, and third-party information updates'",
     "about:package": "echo 'Prepare a package that can then be published on NPM'",
-    "start": "npx ts-node src/main.ts",
+    "start": "npx tsx src/main.ts",
```
donmccurdy, line 88: Note - we can execute TS directly in current Node.js versions. May need to avoid certain features like enums. I switched from tsx in glTF Transform here:

https://github.com/donmccurdy/glTF-Transform/pull/1762

Resolving file extensions/suffixes can be tricky though. Changing to tsx from ts-node may be a good first step in any case.

javagl: I have to admit that, for _some_ aspects of .... all that ..., I usually have a "pragmatic" approach:

- I hope that things keep working when _I do not change anything_ 
  - This should be self-evident, but it certainly isn't...
- when the inevitable happens, and something suddenly does no longer work due to some update from TypeScript 6.0.2 to 6.0.3, then I 🤬 a little, and go through the usual "error-google-stackoverflow/issues"-cycle, trying out the solutions that look promising, until "it works".
  - Yes, I should read and understand the TSConfig documentation, and I should thoroughly and deeply understand what `allowImportingTsExtensions` does, why it has to be set to one value, and why that implies that another property has to be set to another certain value, depending on the `target` and `module` and the node version and the npm version and the TypeScript version, but ... at some point, I just shrug it off. When it works, it works 🤷‍♂️ 

That said: You referred to "current Node.js versions". There's currently only the constraint
```
  "engines": {
    "node": ">=18.0.0"
  },
```
Should this be increased to the latest LTS? 
(I _usually_ try to use "the oldest version that works"...)

javagl:
I think this should largely be done. It currently refers to `gltf-transform` `4.4.0-alpha.1` as a dependency, but that should be OK.

In terms of the changes here: This has been opened nearly exactly one year ago ( 😥 ), with some intermediate state where the extension implementation was _not_ yet part of glTF-Transform. So the process for the update is/was not as "linear" as it could have been if it was started today, but I hope that it's still clear enough. 

Looking at the "Files changed" may be a bit distracting, because the "golden" reference files for the specs have been updated. (This was mostly due to changes in the order of properties in the glTF files). If someone wants to go into the details, looking at the commits may make more sense (skipping ones that just fix typos and such). 

Otherwise, here's a summary of the changes:

- The `(EXT)MeshFeatures` and `(EXT)StructuralMetadata` classes have been removed from the `gltf-extensions` package, and replaced with the ones from glTF-Transform
  - This causes several minor changes in other files, usually only changed import declarations...
- (Aside: The `(EXT)InstanceFeatures` is now the only one left in `gltf-extensions`)
- The `StructuralMetadataMergeUtilities` has been removed: This originally contained functions that had been _taken_ from glTF-Transform because they had not been exported. This has been replaced with the (now exported) functions from glTF-Transform
- A utility at `StructuralMetadataSchemas` has been added for translating between the "JSON-level" and the "glTF-Transform Model-Level" representation of the structural metadata structures.
- The `stringOffsetType` and `arrayOffsetType` properties for structural metdata had been assigned the default value of `UINT32` in the previous extension implementation. In the new implementation, they can now be `string | undefined`. The "Model" implementations for metadata access (in `metadata/binary`) now generally accept `string | undefined` for these values more thoroughly, and only default to `UINT32` "as late as possible"
- Somewhat unrelated to the core changes here: For some reason, `npx ts-node...` calls in the package JSON caused ... some errors. With `npx tsx...`, it seems to work 🤷‍♂️

package.json
donmccurdy: Node.js versions <=20 [no longer receive security updates](https://endoflife.date/nodejs). I usually test against one or two of the active releases (22–26 currently) and ignore everything that's out of maintenance. But using an older version for a test runner sounds pretty harmless, too, not a particular concern!

javagl:
@donmccurdy Not urgent, but unless you have any additional/closing thoughts about this PR, I'd consider this as ready to be merged.

src/tools/contentProcessing/GltfTransform.ts
```diff
@@ -53,12 +51,8 @@ export class GltfTransform {
...
       "meshopt.encoder": MeshoptEncoder,
     });
-    // Note: The order of these calls matters. The EXTMeshFeatures and
-    // EXTInstanceFeatures depend on the EXTStructuralMetadata, because
-    // they may refer to PropertyTable objects via their FeatureId.
-    // So the EXTStructuralMetadata has to be read first.
-    io.registerExtensions([EXTStructuralMetadata]);
-    io.registerExtensions([EXTMeshFeatures]);
+    // Note: The order of the calls matters.
+    // EXTInstanceFeatures depends on the EXTStructuralMetadata
```
donmccurdy: The order of registration shouldn't affect I/O results any longer, and in any case these are now part of the 'ALL_EXTENSIONS' list provided by glTF Transform.

src/tools/gltfExtensionsUtils/StructuralMetadataMerger.ts
```diff
@@ -190,30 +193,34 @@ export class StructuralMetadataMerger {
...
+        targetExtStructuralMetadata,
+        targetSchemaJson
+      );
 
       // If the target schema was resolved from a URI, then assign
       // the newly created object to the target, and set the
       // schemaUri of the target to null
       targetStructuralMetadata.setSchema(targetSchema);
-      targetStructuralMetadata.setSchemaUri(null);
+      targetStructuralMetadata.setSchemaUri(null as any);
```
donmccurdy: Minor - An empty string is what glTF Transform is expecting here, but I'd expect null to work fine too. If an empty string is written to the output file, that would be a bug in glTF Transform.

donmccurdy (approved):
> It currently refers to gltf-transform 4.4.0-alpha.1 as a dependency, but that should be OK.

Stable v4.4.0 version was published this week – I am still considering these two extension implementations experimental, for now.

Beyond that just a couple minor comments, and LGTM!

src/tools/contentProcessing/GltfTransform.ts
javagl: This referred to `EXTInstanceFeatures` in particular, because of that dependency. (It now sounds like it was possible to _first_ register `EXTInstanceFeatures` and _then_ `ALL_EXTENSIONS` - I'll try it out and remove the comment if this is the case)

src/tools/gltfExtensionsUtils/StructuralMetadataMerger.ts
javagl: That was a leftover from the `null`-vs-`""`-confusion from the comment above. Fixed (also in the specs).

PR #194, opened 2025-11-26
Added tool to upgrade legacy 3DGS Tilesets to current spec

weegeekps:
This adds a function to upgrade the legacy 3DGS Tilesets to the latest specification for the 3DGS + SPZ extension. I've added some unit tests, but I have some open questions that I've included in code comments within this PR.

I used the existing upgrade functionality as a guide for how to do this. I decided against using glTF Transform despite my impression that it's preferable these days, as using glTF Pipeline seemed more straightforward and wouldn't require me to finish my roughshod implementation of the `KHR_gaussian_splatting` and `KHR_gaussian_splatting_compression_spz_2` extensions for glTF Transform.

cc: @ggetz & @jjspace since you asked me to complete this.


.gitignore
```diff
@@ -17,4 +17,5 @@ package-lock.json
 /specs/data/migration/output
 /packages/**/build
 *.tsbuildinfo
-3d-tiles-tools-*.tgz
\ No newline at end of file
+3d-tiles-tools-*.tgz
+.idea
```
weegeekps: I use WebStorm for TypeScript & JavaScript so I added this ignore so my project files don't accidentally get checked in.

specs/tools/contentProcessing/GltfUtilitiesSpec.ts
```diff
@@ -148,4 +148,154 @@ describe("GltfUtilities", function () {
     expect(data.jsonData.length).not.toBe(0);
     expect(data.binData).toEqual(Buffer.from([]));
   });
+
+  it("replaceLegacyGaussianSplattingExtensionGltf properly upgrades the legacy 3DGS glTF JSON to the current 3DGS glTF spec", function () {
```
weegeekps: I opted for a "snapshot" style test here since Jasmine is very good about telling you exactly where some JSON isn't matching correctly.

src/tools/contentProcessing/GltfUtilities.ts
```diff
@@ -11,6 +11,38 @@ import { GltfWeb3dQuantizedAttributes } from "./GltfWeb3dQuantizedAttributes";
...
+
+type GltfMesh = {
+  primitives?: GltfPrimitive[];
+};
+
+export type GltfJson = {
+  extensionsUsed?: string[];
+  extensionsRequired?: string[];
+  meshes?: GltfMesh[];
+};
```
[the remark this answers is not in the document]
javagl: I think that things should generally be declared 'as local as possible' (and 'as global as necessary'). Right now, they are only used in the functions of this class. 

As you may have seen, the glTF JSON is usually typed as `any` here. The reason is ... that (there aren't sooo many places where the raw JSON is used after all, and) this is what it is: `JSON.parse` can return `any`-thing, and that `as GltfJson` does not have any notable effect. Given that all properties are `optional?` (even the ones that are required according to the glTF spec, e.g. `mesh.primitives`), there isn't really any added type safety here. (Maybe convenience, like VSCode code completion...?).

In that manner, these types are similar to the types in `src/structure`. And it could be worthwhile to create such types for glTF. This wouldn't have to happen manually - the PR at https://github.com/CesiumGS/wetzel/pull/87 might already be able to work on the glTF schema. (Maybe I'll try it out. It has been a while since I created that PR, but cannot imagine any major hurdles here - maybe just some minor manual tweaks on the output...).

That's only a "drive by comment" (it's late here) - I'll do a proper review soon. 

Before that... maybe a rough estimate: On a scale between 0.0 (just merge it already) and 1.0 (strict!), where should that review be? 

<sub>(You know... I think that this PR introduces the first ... "things" (types, functions, properties) that are _not_ `/** documented */`... )</sub>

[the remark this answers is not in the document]
javagl: Rather out of curiosity, I created a draft PR with the wetzel output at https://github.com/CesiumGS/3d-tiles-tools/pull/195 

I might need more details about the purpose of the "indexer" (assuming that it is the `[...] : unknown` part). 

This seems to work as expected...
```
  const x = {
    componentType: 32
  } as Accessor;
  const c0 = x.componentType; 
  const c1 = x["componentType"];
  console.log(c0);
  console.log(c1);`
```
...and it's not clear for me what the indexer does beyond that.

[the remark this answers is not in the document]
weegeekps: In retrospect it's probably not useful for our cases, but I created a [small repl to demonstrate the concept](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgJICUD2mzIN4BQyxycA-AFzIDOYUoA5gNxEkBGlNdjLAvgQVCRYiFKgAScaqhAATCAA9oyRZDnU0WHPlbEA2gqq16IBsgA+yEAFcAtm2WXqAT3uYANgF0q1kAGsQTAB3ED4BIWh4JDQAOUwZeSUoFQU1WQ0MbFxCEmQETmMeAn4CBEwQWhpMWwgAKWpyn39AkOQAXh1cuCoAIhhsHoAaXWQ2XrY4KCGRhF7sGGnc2V6oODZp3hYCAHpt5AB1YDAAC2QTlFBE6AA6UvKG9whr90wGAAo3hpr68tIMyWkckU0AAlNc4CCWGUKh4ni93p9qnUGiA-mgAQlgVAwWxIXcYY9nq8Pl9kb8pOipJikmCEHjoQ84cTEd8UWiJFSgTS9D0ED1PPT7rCiQjST9URSOYCrtjrrJBQSmaKkeL2RiuaCebJ+XidntDidMNZcOdkJcsbcGcL4SSVWzJXFqaDwQrGSLbazyRlHRrZbioULCTaWWSJd74r7aa7rcyxfbw07sTy+QKA4r3SHVQ6IzKwfK027g3GvbEc1iQVqdSwgA).

When working with JSON it can be beneficial to allow an escape hatch to access _all_ of the potential data on the object, since the use of interfaces actively prevents that. The argument is that since JSON sources are typically not verified, then you should be able to access all the data, even that which is outside of the interface.

Realistically though, when I give deeper thought about it for glTF there's probably only a few specific areas where we would want this. Both `extensions` and `extras` come to mind.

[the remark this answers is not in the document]
javagl: The `extensions` and `extras` _themself_ are already in the `GltfProperty` type, and inherited by all types. (In fact, the "manual cleanup" that I mentioned was to remove exactly these from the _actual_ types: In the glTF schema, they are **re**declared in every type. JSON schema is no real "inheritance" after all...).

This indexer _might_ be justified, given that `additionalProperties` is using the default value of `true` in glTF. But I think that this is rarely used, and offering that indexer to access anything feels a bit like cheating the way around the typing. In standard glTF processing, there should never be a reason to access any non-standard properties. The code that is doing that must _know_ that this property is _there_, in the JSON file that was just read. Whatever the use case for that could be: I'd rather "force" implementors to make that explicit with something like `const p = (value as any).thatProperty;` or so.

src/cli/main.ts
```diff
@@ -552,6 +552,8 @@ async function runCommand(command: string, toolArgs: any, optionArgs: any) {
       targetVersion,
       parsedOptionArgs
     );
+  } else if (command === "upgradeLegacySplats") {
```
javagl: The `combine` command calls `combine`.
The `merge` command calls `merge`.
The `upgrade` command calls `upgrade`.
The `upgradeLegacySplats` command calls `upgradeSplats`.

Not "important", but ... maybe the command should be called `upgradeSplats`...?

More important: Documentation for this new command will have to be added into the main `README.md`, e.g. after the `upgrade` documentation, at https://github.com/CesiumGS/3d-tiles-tools/blob/24bf3b8a90074b5095f51bfaafbb5f2dbf11cf44/README.md?plain=1#L132 

(This may happen at any point in time, and it may make sense to do it "last", after the review. If you had already done it, there would be two places where the command may have to be renamed)

src/tools/contentProcessing/GltfUtilities.ts
```diff
@@ -563,4 +635,122 @@ export class GltfUtilities {
       decodeMatrices
     );
   }
+
+  static replaceLegacyGaussianSplattingExtensionGlb(glbBuffer: Buffer): Buffer {
```
javagl: I'll create a PR on top of this one, with comments for these. 

I know: Some of the comments that I have inserted elsewhere (and the ones that I'll insert here) are "shallow". I won't deny that I'm sometimes ""documenting"" some function like `doSomeComplexOperation` with `/** Does some complex operation */`. These comments are "open-ended", in some way (listing preconditions and postconditions, and providing context). But ... I'll do it. 

(In my local (Java) development, I've configured Eclipse to emit a warning when _anything_ is undocumented, and still maintain these "0 warnings". If someone calls me pedantic and/or claims that this is useless, we can arrange a call with a screen-sharing session, to go through some code (e.g. CesiumJS))

src/tools/contentProcessing/GltfUtilities.ts
```diff
@@ -563,4 +635,122 @@ export class GltfUtilities {
...
+    }
+
+    logger.info("Found legacy 3DGS extension in gltf. Upgrading.");
+    GltfUtilities.replaceLegacyGaussianSplattingExtensionGltf(gltfJson);
+    return Buffer.from(JSON.stringify(gltfJson, null, 2));
+  }
+
+  static replaceLegacyGaussianSplattingExtensionGltf(gltf: GltfJson) {
+    // Update the top level used and required properties.
+    gltf.extensionsUsed?.splice(
```
javagl: For a tileset with `extensionsUsed: [ "KHR_spz_gaussian_splats_compression", "EXAMPLE_extension" ]`, this will clear the array, and create a wrong result. 

The `deleteCount` of `splice` has to be `1` here. 

But... one could argue back and forth about a more "elaborate style"...
```
if (!gltf.extenisionsUsed) { complain(); return; }
if (!gltf.extensionsUsed["...spz"]) { complain(); return; }
gltf.extensionsUsed.splice(..., 1);
...
```
And (in view of the comment about comments), the doc of that function could say
```
/** 
 * ..
 * If the extension is not there, then it complains somehow, and nothing is done 
 */
```

The point is to make sure that the function won't crash or break anything, and to also _say_ that. Callers should not have to fully read (aka reverse engineer) the function, and they should not have to be "afraid" of calling it. The function should be "defensive" and "safe".

src/tools/tilesetProcessing/TilesetSplatUpgrader.ts
```diff
@@ -0,0 +1,293 @@
...
+   * Perform the upgrade of the `Tileset` object, in place.
+   *
+   * @param tileset - The `Tileset` object
+   */
+  async upgradeTileset(tileset: Tileset) {
+    const upgradeRequired = await this.verifyUpgradeRequired(tileset);
+    if (upgradeRequired) {
+      logger.debug(`Upgrading tileset.json`);
+
+      if (typeof tileset.extensions === "undefined") {
```
javagl: Is there is a CesiumJS-style `defined` in
`import { defined } from "../../base";`

In many cases, I've opted for `if (!tileset.extensions)` for checks like this.

(One could argue back and forth. Yes, `if (!tileset.extensions)` would not trigger when `tileset.extensions = "Fooled ya!"`. But this would already violate the type definition of the `extensions` that was defined in `RootProperty`. I think when the _type_ already dictates that it cannot be one of the `""` or `0` cases that are so problematic for `if`-checks in JavaScript, the `if`-check should be OK...)

src/tools/tilesetProcessing/TilesetSplatUpgrader.ts
```diff
@@ -0,0 +1,293 @@
...
+        );
+
+        return;
+      }
+
+      // Update 3DTILES_content_gltf with correct extensions.
+      const upgradedList = [NEW_SPLAT_EXTENSION, NEW_SPZ_EXTENSION];
+      const extensionsRequired =
+        tileset.extensions["3DTILES_content_gltf"]["extensionsRequired"];
+      extensionsRequired.splice(0, extensionsRequired.length, ...upgradedList);
```
javagl: Similar to the other comment about `splice`, this will break when there was a
` extensionsUsed: ["KHR_spz_gaussian_splats_compression", "EXAMPLE_extension"]`
and result in the `EXAMPLE_extension` being omitted in the output. 

This update is, once more, something that can be done in many different ways, with some subjective elements. One could consider using `array.filter(...)` to remove the old one, **but** that doesn't change the array in-place. One could use `splice`, but that requires checking the index (i.e. whether the element was actually contained). It's always more complicated than [`remove`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#remove-java.lang.Object-). One way could be this...
```javascript
      const extensions = tileset.extensions;
      const contentGltfExtension = extensions["3DTILES_content_gltf"];
      if (!contentGltfExtension) {
        // Aaaaaahhh!!!
        return;
      }

      // Comment, maybe make this a `private static` function
      const replaceThem = (extensionsList: string[]) => {
        if (!extensionsList) {
          return;
        }
        const index = extensionsList[OLD_SPLAT_EXTENSION];
        if (index >= 0) {
          extensionsList.splice(index, 1);
        }
        extensionsList.push(NEW_SPLAT_EXTENSION);
        extensionsList.push(NEW_SPZ_EXTENSION);
      };
      replaceThem(contentGltfExtension.extensionsUsed);
      replaceThem(contentGltfExtension.extensionsRequired);
```
But I don't have a strong preference about some details, **beyond** that it should (safely) do the right thing.

src/tools/tilesetProcessing/TilesetSplatUpgrader.ts
```diff
@@ -0,0 +1,293 @@
...
+   */
+  private async verifyUpgradeRequired(tileset: Tileset): Promise<boolean> {
+    if (!Extensions.containsExtension(tileset, "3DTILES_content_gltf")) {
+      logger.error(
+        "3DTILES_content_gltf extension not found. Are you sure this is a 3DGS tileset? Unable to continue."
+      );
+      return false;
+    }
+
+    if (typeof tileset.extensions === "undefined") {
```
javagl: See other comment about `undefined` check

javagl (commented):
>> On a scale between 0.0 (just merge it already) and 1.0 (strict!), where should that review be?
>
> On that scale, I'd say maybe a 0.5?

I might have gone into the ">0.75 range" with some of the inlined comments. 

Higher-level:

Most of the other `...Upgrader` specs use the "golden" approach, actually upgrading a tileset that is stored as a file in the `Specs/data` directory and comparing the result to the expected ("golden") one. But given that the upgrades here only twiddle with the JSON (both tileset and glTF), this may not be necessary.

It would still be nice to have a "full, real-world" test. Some tileset with something like the "unit cube" that was _actually_ generated by the tiler (based on a PLY file), and used _verbatim_ here, verified to be 1. rendered properly by CesiumJS in its old version, and 2. rendered _exactly the same_ by CesiumJS in its new version. 

But ... it's difficult to add something like this _now_. It would have been neceessary to diligently track the outputs of actual unit (cube) tests from the tilers (with their exact versions and changes due to bugfixes), and align that with the CesiumJS changes and bugfixes. 

I'll probably _try_ to find/download some data from the tilers that was generated with the old extension (I might have some in my ion account), and apply the upgrade to this data to see how it goes. If anything doesn't work there, it will be impossible to say for sure whether it's a tilers issue, an upgrade issue, or a CesiumJS issue. But it could be worthwhile to at least have one "snapshot result", to validate this with some _real_ data.

(Or... did you already try this out with real, old ion data?)

weegeekps:
> You know, the tilers might have written some "up-axis transform matrix" into the glTF as a node.matrix in one version, but applied some coordinate correction directly in the splats in another version or so.

I hope not because @keyboardspecialist and I wrote all that code, and we tried to keep it consistent between versions. There were many very late caffeine-fueled nights there near the end of the first version, so I guess it's possible. 😬

Let me know if you want the SHAs for the tiler versions to produce any test data. Since this repo is public, I will send those to you directly if you want them.

src/tools/contentProcessing/GltfUtilities.ts
javagl: I just realized that much of the used/required handling should be covered by the [`Extensions`](https://github.com/CesiumGS/3d-tiles-tools/blob/c027ac64849794d70703807c5981ca0fa0c22e60/src/tilesets/tilesets/Extensions.ts#L61) utility class. So there could be code like
```
if (Extensions.usesExtension(gltf, "...SPZ")) { ... }

Extensions.removeExtensionUsed(gltf, "OLD_SPZ");
Extensions.addExtensionUsed(gltf, "NEW_SPZ");
 ```
where the latter functions already ensure consistency.

PR #208, opened 2026-07-21
Add CESIUM_primitive_outline support

### LINKED ISSUE #207: Missing optional extension, "CESIUM_primitive_outline"
When upgrading `.b3dm` files to 3D Tiles 1.1, I receive the following warning:

```text
Missing optional extension, "CESIUM_primitive_outline".
```

Does `3d-tiles-tools upgrade` currently support tilesets containing the `CESIUM_primitive_outline` extension?

If not, is there a recommended way to preserve the outline data when converting legacy `.b3dm` content to 3D Tiles 1.1?

javagl:
Fixes https://github.com/CesiumGS/3d-tiles-tools/issues/207  

This adds support for the [`CESIUM_primitive_outline`](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/CESIUM_primitive_outline/README.md) extension, based on glTF-Transform (cc @donmccurdy )

Given that the extension only defnies a single object that is attached to mesh primitives and defines an `indices` accessor that contains the outline indices, the implementation seems to be pretty straightforward. 

The only thing that I wasn't entirely sure about was the 'usage' for the accessor
`return this.setRef("indices", indices, { usage: "OTHER" });`
`                                                ^ ???      `
but I assume that 'OTHER' is OK here. 

EDITED: 
The 'OTHER' was commented on, but it seems like 'OTHER' is indeed a sensible choice for now.

I also added a basic spec based on the GLB version of https://github.com/CesiumGS/3d-tiles-samples/tree/main/glTF/CESIUM_primitive_outline/BoxPrimitiveOutline , just reading the file and checking the outline indices against the expected ones.


src/gltf-extensions/gltfExtensions/PrimitiveOutline.ts
```diff
@@ -0,0 +1,40 @@
...
+  protected override getDefaults() {
+    return Object.assign(super.getDefaults());
+  }
+
+  public getIndices(): Accessor | null {
+    return this.getRef("indices");
+  }
+
+  public setIndices(indices: Accessor): this {
+    return this.setRef("indices", indices, { usage: "OTHER" });
```
donmccurdy: I believe the `WriterContext.BufferViewUsage.OTHER` enum value is accessible here.

src/gltf-extensions/gltfExtensions/CESIUMPrimitiveOutline.ts
```diff
@@ -0,0 +1,107 @@
...
+    primitiveDef: GLTF.IMeshPrimitive
+  ) {
+    const primitiveOutline = primitive.getExtension<PrimitiveOutline>(NAME);
+    if (!primitiveOutline) {
+      return;
+    }
+    const indices = primitiveOutline.getIndices();
+    if (!indices) {
+      return;
+    }
```
donmccurdy, line 98: Is the `indices` property required  when this extension is present? It doesn't appear so from the definition, but it's also not clear what an implementation can/should do without indices. Just wondering if a silent return is the desired behavior here, similar to above.

src/gltf-extensions/gltfExtensions/CESIUMPrimitiveOutline.ts
```diff
@@ -0,0 +1,107 @@
...
+    const meshDefs = jsonDoc.json.meshes;
+    if (!meshDefs) {
+      return this;
+    }
+
+    for (const mesh of this.document.getRoot().listMeshes()) {
+      const meshIndex = context.meshIndexMap.get(mesh);
+      if (meshIndex === undefined) {
+        continue;
+      }
```
donmccurdy: Nit/optional: This would be an invalid state in the data model, typically in glTF Transform I'd be tempted to let it reach a runtime error, either with non-nullable assertion (`!`) or `assert(condition, msg)`.

donmccurdy (approved):

src/gltf-extensions/gltfExtensions/PrimitiveOutline.ts
donmccurdy: You might get better meshopt compression results if meshopt knows to consider this more like `ELEMENT_ARRAY_BUFFER`, but I suspect glTF Transform's implementation will need more hints than just the usage to know what to do with this accessor. Probably would not worry about that in this PR.

javagl:
Thanks @donmccurdy for the review (I didn't expect this, but of course, appreciate the input). I'll address this ASAP (maybe waiting if there are additional comments in the issue or here).

Two of the inlined comments have been about ~"error handling". This also came up during the implementation of the metadata extensions, and I wasn't sure about the best practices. 

On the one hand, one can _very_ reasonably make the assumption that the input file is _valid_, and the "undefined behavior" for invalid input may be arbitrary, including a plain crash or error being thrown. (And I've seen that glTF-Transform generally handles these cases with `throw Error`)

On the other hand, I'm usually trying to be somewhat lenient: Processing 999999 tiles of a tileset and then brutally bailing out with an error because the 1000000ths one was missing some JSON property can be a nuisance. (Particularly, because this specific property is not _actually_ required for reading or processing the other data - it is only technically 'required' in the context of the extension spec...)

In any case, you're rightfully pointing that out, because the current behavior is "fail silent", and that's even worse. I'll think this through. It will probably be at least a `logger.warn("Ignoring [...that invalid stuff...]");`.

javagl:
I assume that the invariant
```typescript
const meshes = document.getRoot().listMeshes();
const mesh = meshes[indexInArray]; 
const indexInMap = context.meshIndexMap.get(mesh);
assert (indexInArray === indexInMap); // !?!
```
holds. (I.e. I assume that the index map is only intended for the case where the index of some random `mesh` actually has to be looked up, and it's a map to avoid doing that O(n) search). I now changed this to a plain `for` loop.

For the other error case, I added a `logger.warn` output. One _could_ simply assume valid glTFs, but given that `CESIUM_primitive_outline` is not (and likely will never be) supported by the validator, handling this unlikely case with a 'warning' seems reasonable.

src/gltf-extensions/gltfExtensions/CESIUMPrimitiveOutline.ts
javagl: Added a `logger.warn` for this case (see comment below)

src/gltf-extensions/gltfExtensions/CESIUMPrimitiveOutline.ts
javagl: I changed this to a `for` loop now, which avoids the index lookup to begin with. (See comment below)

javagl:
The last commits consisted mainly of minor "chores", and the core has been approved, so I'll just go ahead and merge this one.

